PR #13, opened 2016-02-12
First attempt at using maven

mjwall:
run 'mvn clean package' from the root then run
'java -jar assembly/target/bsh-2.1.0-SNAPSHOT.jar'
to see the built jar work

Not really looking for this to be merged.  Just using github PRs as a tool for review and discussion.  This is a little different from the maven branch that was pushed up here, which I didn't notice until I started writing up this PR.  Not sure I got as far and saw some answers to my questions from the README.maven in my branch.

If this is useful, I can work on it some more.

Thanks


mjwall:
Closing now that I have had a chance to look at the maven branch.  Much cleaner than what I did.

PR #14, opened 2016-02-15
Maven executable jar

mjwall:
Add mainClass to manifest of jar so you can run

java -jar target/bsh-2.1.0-SNAPSHOT.jar

Also cleaned up some IDE files, in a separate commit that can be removed if desired.

@stain let me know if this is useful.


mjwall:
grr, should have been against maven branch

PR #15, opened 2016-02-15
Maven executable jar

mjwall:
Add mainClass to manifest of jar so you can run

java -jar target/bsh-2.1.0-SNAPSHOT.jar

Also cleaned up some IDE files, in a separate commit that can be removed if desired.

@stain let me know if this is useful.


stain:
Thanks, @mjwall ! I've merged this into the `maven` branch -- in your opinion do you think the `maven` branch should be ready to be merged into `master`? It seems to work for me.. 

BTW, I have invited you to the beanshell GitHub project. :)

mjwall:
@stain, thanks for the invite.  I don't know if the maven branch builds all the documentation like the ant script used to.  I also feel like we can get the javacc stuff working.

stain:
No, it doesn't build the documentation, but I moved it to https://github.com/beanshell/beanshell/wiki - so I removed the doc/ folder from the maven branch now.

I think the only thing missing from the wiki is the autogenerated bshdoc thingies - but I can't get that to work with ant either.

On naster, `ant bshdoc` generates `docs/manual/bshcommands-bshdoc.xml` - but if I then try to build it inside docs/manual with `ant` I get:

```
java.lang.RuntimeException: Unrecognized XSLTC extension 'org.apache.xalan.xslt.extensions.Redirect:write'
```

However I managed to move it into https://github.com/beanshell/beanshell/wiki/Commands using Firefox's XSLT instead.

PR #19, opened 2016-02-19
18 source release

### LINKED ISSUE #18: Source release as part of Maven build
The earlier ant build also prepared the source distribution zip, e.g. http://central.maven.org/maven2/org/apache-extras/beanshell/bsh/2.0b6/bsh-2.0b6-src.zip

This is easy to achieve with MAven as well, e.g. as done by the `apache-release` in the [ASF super pom](http://central.maven.org/maven2/org/apache/apache/17/apache-17.pom)

This bug is to update our pom.xml to do something similar (we can't simply use Apache's super-pom as we're not an ASF project).

mjwall:
Build a source jar 

Attach to the verify lifecycle so it runs when you do mvn install or mvn deploy
but not when you run mvn package.  That can be changed if needed.

Should just go up on a mvn deploy.

Also formatted the pom in a separate commit that can be removed. 

closes #18

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,195 +26,195 @@
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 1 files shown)


nickl-:
I am with you on the source formatting, took me more than an hour to cherry pick through all the whitespace crud. Before we work on this again we should format all the source to conform to some style.

PR #20, opened 2016-03-12
improve compatibility with different Invocable implementations

dougmencken:


nickl-:
Nice one!

stain:
@dougmencken -- would you mind having a rebase and try again? It seems like you did a bit of re-indentation which caused this now to conflict in TestBshScriptEngine.java

dougmencken:
Pushed an updated version. The only change was addition of "package bsh;" on top of that file

PR #21, opened 2016-04-12
Replacing deprecated junit.framework.Assert with org.junit.Assert

MrEasy:


nickl-:
I got these in #36 too...

PR #39, opened 2017-09-05
improve compatibility with different Invocable implementations

dougmencken:


nickl-:
Do you have documentation to support your claim?

Implementation looks ok but you can keep the whitespace.... I agree the whitespace is a headache but we should fix it all not just the one document.

We fixed travis now, if you resubmit it should pass the build now.

nickl-:
Please re roll new PR against default branch.

PR #45, opened 2017-09-21
Added varargs support

jaktao:
I have added the varargs support. I have done this merge using bsh.Reflect.java into this fork https://github.com/pejobo/beanshell2 . It is possible to call the method with varargs and not. For example if you have foo(String... values) , now you can call foo("A", "B", "C") or foo(new String[]{"A", "B", "C"}). Please check and test before and please add this feature into this fork. I would like use this fork.

### DIFF
--- a/src/main/java/bsh/Reflect.java
+++ b/src/main/java/bsh/Reflect.java
@@ -27,8 +27,17 @@
 
 package bsh;
 
-import java.lang.reflect.*;
-import java.util.Vector;
+import java.lang.reflect.Array;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Member;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.ArrayList;
 
 /**
  * All of the reflection API code lies here.  It is in the form of static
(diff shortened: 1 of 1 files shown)


nickl-:
Damn man that bloody whitespace!!! I can't see your changes between all the crud =(

I see the comments are a hassle for you too... been working on the code style but there doesn't seem any way to format the comments without adding the additional *'s in the block comments. 

Normally that wouldn't be an issue right... but beanshell parses it's own documentation with a bsh script so I need to make sure to check if that will still work after the code clean-up. Almost done just need a few more hours without distraction.  

Unfortunately I can't merge if i don't know what changed. If you can submit a patch with only your changes, without disturbing the whitespace. I can have a look at it but otherwise we'll have to wait for the cade cleanup to be done so we can put these messy pull requests behind us.

O wow, ty for pointing out beanshell2, looks like these guys at  code.google.com/p/beanshell2 forked us at some point to be able to make changes. I am interested to see what they've managed to add, hopefully it will merge easily. Looks like their project came to an end too... 

I'll open another issue for that... tx!!

jaktao:
Hi @nickl thanks for reply. I know the problem of code style, in this case i formatted the original code and code based on [pejobo/beanshell2](https://github.com/pejobo/beanshell2) branch only for Reflect.java file. 

Attachments: [Reflect.java.zip](https://github.com/beanshell/beanshell/files/1351622/Reflect.java.zip)

I don't know how it is possible to do "pull request" without create a copy of this branch.

nickl-:
You don't have to make a copy, that will mean you need to create a new PR as this one is specific to this branch. All you do is fix this tree, you could make changes and additional commits or rebase and amend the commits that are here already. It will require you to force push but GitHub handles that without complaints. 

But don't sweat it, I've actually already ported this and might've already pushed it too if I recall... let me see if I can find it quick. Ahh yes I think this was the one beanshell/beanshell@bca5d33 ... am I correct?

I was actually under the assumption that varargs support meant we could declare varargs in script but alas it only means beanshell can now call vararg methods in Java, which also needs to happen don't get me wrong. I couldn't quite figure out how they managed to implement this change without modifying the grammar lol... wishful thinking.

nickl-:
We now have full varargs support see beanshell/beanshell@c1adf5d

Closing...

stefanofornari:
oh wow! now I really need to switch to v2 :)


Ste

PR #47, opened 2017-09-24
Clean style

nickl-:
To simplify patch/pull request process we needed to conform the code to a single style and normalise whitespace. By applying style validation to the build chain we ensure that patches submitted will already be clean. This will make life much easier for both contributor and maintainer as the patches will only contain the mods.

The shortest route to accomplish this was to apply the default checkstyle, being the [sun java code style](http://www.oracle.com/technetwork/java/codeconvtoc-136057.html), and configuring exceptions to the rule instead of defining a new set of rules from the ground up.

### What is the rule?

In short the convention is using only the space as white space, 80 character page width, 4 spaces indent, block opening inline and closing alone on new line. It also expects javadoc style comments applied in full. 

### What are exceptions to the rule?

 *  Allow TODO: The rule set disallows TODO comments, which we find useful. 
 *  Relax on page width: Even though we are basically within the 80 character width there are a few occurrences where we go slightly over the max and forcing the width will not be an improvement. 
 * Allow single line statements: The standard is obsessed with structures, even require array definitions to be written as a block, we relax those rules to accommodate:
    *  Allow single line array definitions

              String[] foo = new String[] { "bar", "baz" };

    *  Allow inline conditionals

             boolean may_i = condition ? truthy : falsey;

    * Allow one line statements:

            while (true)
                for (item : list)
                    if (isValid(item))
                          item.apply(style);
                   else
                         break;

    * Allow empty blocks on 1 line:

            public void methodNotImplemented(String ignored) {}

 * Relax on declarations: The rule requires not declaring modifiers in interfaces, public is considered redundant in JDK 8 but Java 9 introduces private interface members so why shouldn't we be able to stipulate our intent. The fact that there is no reliable means to impose this rule on an existing code base, eclipse et. al. are incapable to apply this mod, is completely besides the point. 
 * In addition to style the standard also imposes restrictions where common problems may arise with regards to software design and coding practices. These along with naming conventions are beyond the scope of this patch as they require refactoring.  Currently also marked as disabled but once we have improved our test coverage we can look at enabling them again.

The same results is achieved with with the current test suite after applying the patch which proves (to modest degree) that the changes are all cosmetic and no functionality has been modified.

@stain will give you a few days to respond or assume you are happy to have it merged to HEAD otherwise.


nickl-:
Closed in favour of #48 for now.

PR #48, opened 2017-10-01
Clean style limited

nickl-:
In favour of #47 until we have finished merging with [pejobo/beanshell2](pejobo/beanshell2) and are back at generating our grammar.

It was still ok to merge with the fork but when I looked into updating the grammar to include newer features in java things went sideways and it was not possible to spot the differences we might have made to the generated source back then.

This now does the bare minimum cleanup:
 * Tabs to 4 spaces
 * No trailing white space
 * CRLF to LF (dos2unix)
 * NL at end of file

Which should make it easier to submit patches, we'll have to wait and see.

Also applied check style as before with even more exclusions now but this will ensure the source does not further degrade as we continue.

This patch includes a commit to fix imports:
 * Flatten * imports
 * Remove unused imports 

Which were flagged by checkstyle...


nickl-:
Closed: Already applied.

PR #49, opened 2017-10-01
Upgrade asm to 1.5.3.

nickl-:
This is currently the last of the 1.x branch. It includes Attributes but allows us to pass null
or the attribute when writing and I needed to comment out the read method as we don't include
the ClassReader.

Tried to upgrade to 2.x but did not go well, we will have to change our approach to the preferred
method if we want to upgrade further.

Assumes #48


nickl-:
Already merged with branch merge-fork-beanshell2... closing... and will delete branch

PR #75, opened 2018-06-24
Lightwaight parser-reset

stefanofornari:
In order to discard input entered on ^C i need to reset the parser and underlying input stream. This patch proposes an easy method that can then be used by subclasses of Interpreter.

Somehow related to https://github.com/stefanofornari/BshConsole/issues/2


nickl-:
We don't need this anymore... do we?

stefanofornari:
I would appreciate we include it as per the other comments.... :)

Ste

nickl-:
I really tried with but it doesn't want to work for me.

When I call `resetParser` straight out of the blocks, `in` is close which triggers an EOF and interpreter promptly exits the VM.

I can of course avoid the `System.exit();` call by turning exit on EOF off `bsh.setExitOnEOF(false);` but closing the input still triggers an EOF and Interpreter abandons the REPL loop. Now what?

There are several ways... including the autoclosable contract method,`close()`. For closing the input reference held by the interpreter but it can also be closed outside. I could very likely be persuaded that a simpler method for resetting or interrupting a running interpreter is beneficial.

Currently the steps include: (not necessarily in order)

 * Interrupt running thread
 * turn exit on EOF off (very very NB or there's not much point in ever trying any of this)
 * close input and trigger EOF
 * initialise new input
 * initialise new interpreter (adding old interpreter as parent to retain state)
 * start new thread REPL

But it just seems awkward to expect Interpreter to handle this for itself, doesn't it? Perhaps a Runner or similar helper would feel more natural, I just can't help but feel this functionality resides externally. 
 
Also the additional changes that crept into this PR distract from the topic. I don't know how I feel about the Token manager catch block but this is not the place to discuss that either and the jacoco report has its own PR already. 

I actually think it would be easier to just close this PR and start with a fresh branch and new PR if we are going to take this further. Trying to fix this might be more trouble than it is worth.

STATUS: Needs work

stefanofornari:
Hi nick,

sorry for the confusion with the PR. You can close it. I moved this
discussion to the bshconsole ticket.

Ste

nickl-:
Closed: as suggested

PR #79, opened 2018-07-01
Prompt management in console

stefanofornari:
This PR is to delegate the management of the prompt to the the ConsoleInterface. This allows a better integration with console libraries which usually manage the prompt on their own. It seems to be a functionality closer to the responsibility of the console rather than the interpreter.

It fundamentally adds a prompt() method to ConsoleInterface so that when Interpreter needs to show the prompt it calls the ConsoleInterface method. The code will probably explain it better than my words :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -296,6 +296,12 @@
       <scope>provided</scope>
       <optional>true</optional>
     </dependency>
-
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+      <version>1.16.1</version>
+      <scope>test</scope>
+      <type>jar</type>
+    </dependency>
   </dependencies>
 </project>
--- a/src/main/java/bsh/ConsoleInterface.java
+++ b/src/main/java/bsh/ConsoleInterface.java
@@ -43,5 +43,6 @@ public interface ConsoleInterface {
     public void println( Object o );
     public void print( Object o );
     public void error( Object o );
+    public void prompt( String prompt ); 
 }
 
--- a/src/main/java/bsh/Interpreter.java
+++ b/src/main/java/bsh/Interpreter.java
@@ -484,7 +484,7 @@ public class Interpreter
                 Thread.yield();  // this helps a little
 
                 if ( interactive )
-                    print( getBshPrompt() );
+                    prompt(getBshPrompt());
 
                 eof = Line();
 
@@ -1292,6 +1292,11 @@ public class Interpreter
 
     public static boolean getSaveClasses()  {
         return getSaveClassesDir() != null && !getSaveClassesDir().isEmpty();
+    }
+
+    @Override
+    public void prompt(String prompt) {
+        print(prompt);
     }
 }
 
--- a/src/main/java/bsh/util/AWTConsole.java
+++ b/src/main/java/bsh/util/AWTConsole.java
@@ -25,6 +25,8 @@
  *****************************************************************************/
 package bsh.util;
 
+import bsh.ConsoleInterface;
+import bsh.Interpreter;
 import java.awt.Color;
 import java.awt.Font;
 import java.awt.Frame;
@@ -43,9 +45,6 @@ import java.io.PipedOutputStream;
 import java.io.PrintStream;
 import java.io.Reader;
 import java.util.Vector;
-
-import bsh.ConsoleInterface;
-import bsh.Interpreter;
 
 /*
     This should go away eventually...  Native AWT sucks.
@@ -350,4 +349,9 @@ public class AWTConsole extends TextArea
     // unused
     public void keyTyped(KeyEvent e) { }
     public void keyReleased(KeyEvent e) { }
+
+    @Override
+    public void prompt(String prompt) {
+        print(prompt);
+    }
 }
--- a/src/main/java/bsh/util/JConsole.java
+++ b/src/main/java/bsh/util/JConsole.java
@@ -49,7 +49,6 @@ import java.io.PipedOutputStream;
 import java.io.PrintStream;
 import java.io.Reader;
 import java.util.Vector;
-
 import javax.swing.Icon;
 import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
@@ -760,6 +759,11 @@ public class JConsole extends JScrollPane
         } else {
             run.run();
         }
+    }
+
+    @Override
+    public void prompt(String prompt) {
+        print(prompt);
     }
 
     /**
--- a/src/test/java/bsh/InterpreterConsoleTest.java
+++ b/src/test/java/bsh/InterpreterConsoleTest.java
@@ -1,0 +1,175 @@
+/*****************************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one                *
+ * or more contributor license agreements.  See the NOTICE file              *
+ * distributed with this work for additional information                     *
+ * regarding copyright ownership.  The ASF licenses this file                *
+ * to you under the Apache License, Version 2.0 (the                         *
+ * "License"); you may not use this file except in compliance                *
+ * with the License.  You may obtain a copy of the License at                *
+ *                                                                           *
+ *     http://www.apache.org/licenses/LICENSE-2.0                            *
+ *                                                                           *
+ * Unless required by applicable law or agreed to in writing,                *
+ * software distributed under the License is distributed on an               *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY                    *
+ * KIND, either express or implied.  See the License for the                 *
+ * specific language governing permissions and limitations                   *
+ * under the License.                                                        *
+ *                                                                           *
+/****************************************************************************/
+
+package bsh;
+
+import java.io.ByteArrayOutputStream;
+import java.io.InputStreamReader;
+import java.io.PrintStream;
+import java.io.Reader;
+import java.io.StringReader;
+import static org.junit.Assert.assertEquals;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+
+public class InterpreterConsoleTest {
+
+    @Rule
+    public final ExpectedSystemExit exit = ExpectedSystemExit.none();
+
+
+    /**
+     * <a href="http://code.google.com/p/beanshell2/issues/detail?id=50">Issue #50</a>
+     */
+
+    @Test
+    public void read_from_provieded_reader() throws Exception {
+        final StringReader r = new StringReader("var = \"hello\";");
+
+        exit.expectSystemExit();
+
+        Interpreter bsh = new Interpreter(givenConsole(r));
+        bsh.run();
+
+        assertEquals("hello", bsh.get("var"));
+    }
+
+    @Test
+    public void write_to_provieded_printer() throws Exception {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+        final StringReader r = new StringReader("print(\"hello\");");
+        final PrintStream p = new PrintStream(baos);
+
+        exit.expectSystemExit();
+
+        Interpreter bsh = new Interpreter(givenConsole(r, p));
+        bsh.run();
+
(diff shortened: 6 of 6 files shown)


nickl-:
@stefanofornari  Thank you for the PR!

STATUS: Can't merge: Needs work:

 1. CheckStyle fails the CI build (see travis).

There is a bash script located at [scripts/bsh-code-style-fixup](https://github.com/beanshell/beanshell/blob/merge-fork-beanshell2/scripts/bsh-code-style-fixup) which can help with fixing the white space issues.

 2. Try and avoid unnecessarily adding additional dependencies.

you can use:

```java
bsh.setExitOnEOF(false);
```

to prevent `Interpreter.class` from calling `System.exit();` on `EOF` this will save us from having to add the new dependency. Besides I vaguely remember an existing junit rule which should accomplish the same... unless I am mistaken. 

Anyway new dependencies means more work, we need to maintain these dependencies, keeping them updated, checked for vulnerabilities, hope they keep backward compatibility, etc etc etc. The benefit should far outweigh the overhead... please motivate.

 3. Code freeze

Or maybe just a code chill =) we all know there is a lot still to be done on BeanShell but we are trying to work towards a stable release. New features are acceptable as long as they address an existing issue. 

It is not clear what problem the addition of the `prompt()` method aims to resolve. It seems to only wrap the `print` method, so why don't we just stick with print for now? Rather work towards code coverage instead of adding new lines of code which seems counter productive. This topic like all others remain open for discussion, of course.

 4. Prevent tests from outputting to stdout or stderr. 

It can be as simple as replacing `System.out` with a custom print stream using `System.setOut`. With Interpreter there are several other options for replacing the output stream in use. You can also look at `Interpeter.redirectOutputToFile(filename)`, this file can then be used for asserts again. Do remember to also delete this on clean up.

stefanofornari:
>     CheckStyle fails the CI build (see travis).
> 
> There is a bash script located at scripts/bsh-code-style-fixup which can help with fixing the white space issues.

ups, sorry, I removed the check in my local pom because it it really annoying... :) I will resubmit if you are ok with the approach.


>     Try and avoid unnecessarily adding additional dependencies.
> 
> you can use:
> 
> bsh.setExitOnEOF(false);
> 
> to prevent Interpreter.class from calling System.exit(); on EOF this will save us from having to add the new dependency. Besides I vaguely remember an existing junit rule which should accomplish the same... unless I am mistaken.

I guess here the definition of "unnecessarily" is key :) With the trick you propose, you are testing when the class does not call exit(). And how would you test when the class does call exit()??? and what if you have all your tests with the trick and then you go in production without? how can you rely on your code?
I am really a TDD oriented (even if it is not very clear in bshconsole because it is still an experiment) and I love to work with production code. I put a trick only if I do not have any other choice ;)

> Anyway new dependencies means more work, we need to maintain these dependencies, keeping them updated, checked for vulnerabilities, hope they keep backward compatibility, etc etc etc. The benefit should far outweigh the overhead... please motivate.

See above for the motivation. TDD is must to me and any tool I can use to make it easier is a gem. I do not have the same experience as you with maintenance cost of these tools. I have to say that my experience is they are 0-maintenance (while a bug due to I did not know the behavior when exit() is called is definitely higher maintenance :) ).


>     Code freeze
> 
> Or maybe just a code chill =) we all know there is a lot still to be done on BeanShell but we are trying to work towards a stable release. New features are acceptable as long as they address an existing issue.
> 
> It is not clear what problem the addition of the prompt() method aims to resolve. It seems to only wrap the print method, so why don't we just stick with print for now? Rather work towards code coverage instead of adding new lines of code which seems counter productive. This topic like all others remain open for discussion, of course.

Well, it is your call indeed. My point of view is to use beanshell to build a console (that then I use to do data analysis). The two requirements "use the snapshot version" AND "let's do not change too much the snapshot" do not go along too well :) you tell me which one you prefer.

Introducing prompt() makes it easier to build a console because Interpreter should not really manage the prompt. It is currently broken (IMHO). Indeed looking at beanshell only, I am just renaming a print() into a prompt(), but if you look at it from who is developing an interactive console, you should see the difference. In the latter case the console is notified of a specific event "please show the prompt". I may agree though this can still be considered a trick and we need the interpreter to generate events...

> 
>     Prevent tests from outputting to stdout or stderr.
> 
> It can be as simple as replacing System.out with a custom print stream using System.setOut. With Interpreter there are several other options for replacing the output stream in use. You can also look at Interpeter.redirectOutputToFile(filename), this file can then be used for asserts again. Do remember to also delete this on clean up.

I guess the counter-argument is just in the workaround you are suggesting... do not reinvent the wheel with a workaround you then have to remember you did... come to the dark side of system-rules (and maybe one day fluent assertj assertions), you won't go back! :)

PS: sorry for the mess with the PR, am completely new to this...

nickl-:
> On  definition of "unnecessarily" vs proposing a trick to circumvent default application flow

I would like to separate these two arguments and point out that if you need a special dependency in order to "avoid a trick" are you confident that it isn't a "trick" by itself which is now circumventing default application flow. In other words, I am not convinced this is a valid argument to motivate that a dependency is necessary, how can it be normal behaviour without the dependency? Isn't this what you aim to test for?

> I am really a TDD oriented and I love to work with production code. I put a trick only if I do not have any other choice

I agree with you completely and I wouldn't ever suggest a "trick" as a work around to avoid testing a normal path of execution. This is more an argument for why are we testing `System.exit` or more specifically is it arguably "untestable" because it  prevents a test from completing which is a requirement for a test to succeed. 

We won't go into the debate on whether `System.exit` is good coding practice or not. What is more significant here is the fact that by capturing or preventing the `System.exit` is in itself circumventing the normal process flow of your application. The default behaviour was for the VM to die which was now prevented.

My suggestion was not really a trick or a mock, I would argue catching an exit is a trick, what we have with "set exit on POE" is another "default" path that has more reasons to be thoroughly tested.

> a bug due to I did not know the behavior when exit() is called

I would challenge you to find me such a bug, there is not much left to go wrong once we've pulled the plug. I am however willing to bet good money that you are more likely to find a bug which is only relevant to the scenario you introduced by capturing or preventing `System.exit` from completing. 

A Bug by definition is something happening other than what was expected, not much expectation can remain beyond a call to `System.exit` unless we didn't actually exit.

>  I do not have the same experience as you with maintenance cost of these tools.

There is very little cost in maintaining your own source code and I am not referring to the cost of upkeep of currently maintained code. It might be obvious in something you haven't touched in years.

This becomes more relevant in scenarios where code was allowed to get outdated or left unmaintained and these dependencies were introduced by other developers which are no longer available. This is more often the case with open source projects but should be a consideration when you are paid to leave source code which someone else is likely to maintain.

You will know exactly the cost I am referring to next time you take on someone else's code.

You are likely to wish there were less dependencies rather than hoping they had more. The most obvious dependencies will be easier to maintain and have a larger impact on the code base (If you broke something it is obvious). It is those edge cases, nice to haves, less often used requirements which will take up more of your time. Now ask yourself, was it really "necessary" to add them in the first place and the answer is HELL NO. That is all I am saying =)

Why we care to develop code that is easier for others to maintain is more of a philosophical question.

nickl-:
> PS: sorry for the mess with the PR, am completely new to this...

No worries, we were all new at this at some point, glad you said I am happy to help you along. First and foremost there is absolutely no need to apologise for submitting a PR... ever! Your only concern should be how badly you want your changes to be applied, which will be an indicator for how much effort you are willing to put into that patch.

What is happening here is what we call "working together on a PR or change", even though it may feel like you are the only one that seems expected to do any work =) Basically there are two roles, the owner or contributor is the person who will be editing the code and the reviewer, committer or maintainer is usually the person that will commit the code. There may be multiple parties for each role.

I like to be more proactive and tend to also supply some possible solutions you could try but others may only give you criticism where it is up to you to find a resolve. There is no reason to feel bad about a PR no matter how many times you are asked to make changes. Obviously you won't bother to make fixes if you didn't care about the change but it helps to keep in mind the other person wouldn't bother to work on it with you if they didn't feel there is value. So basically no one owes anyone anything but least of all an apology. Try and look at any response in a positive light... it far outweighs no response at all.

Don't let it bother you if it's not perfect the first time, it is far more important to make yourself available and respond to requests in a timely fashion. Open communication is best so make it clear to the other parties how much time you have to work on it.

You are on the right track, create a new branch for every PR you submit. Any changes you make can simply be pushed to that branch on your fork and it will automatically update the PR. You can also start over without changing the PR. Delete your local branch and recreate it, make the new changes and commits and then push it to your GitHub fork branch with the `--force` argument. This will discard the older changes and replace it with the new ones. The PR should automatically update.

Shout with any questions...I am here to help.

stefanofornari:
thanks nick. what about the proposed changes? I need to use bshconsole and
therefore to know which code base I should take into acocunt.

Ste

nickl-:
>  I need to use bshconsole and therefore to know which code base I should take into acocunt.

the only BeanShell code base I will recommend at this point for any circumstances is HEAD of merge-fork-beanshell2

make sure it is HEAD of merge-fork-beanshell2 at any given time, to ensure you have the latest updates.

nickl-:
> what about the proposed changes? 

The status of this PR is still

STATUS: Needs Work

I gave you some pointers to consider. It is entirely up to you wether the changes get committed.

stefanofornari:
HI nick,
thanks for the feedback. I am still not clear if given I will fix the style
thing the change is acceptable. I provided by explanation and feedback.
I do not know if we are making much progress. Maybe the best is that I just
write my own version of Interpreter and do not try to change beanshell.

thanks a lot for your effort and help.


Ste

nickl-:
@stefanofornari This is an anti climax, was looking forward to working with you.

As you wish...hope this does not discourage you from sending subsequent contributions though.

nickl-:
Opened: I would still like us to work on this and get it committed, please.

stefanofornari:
sounds good. now that we have a bit more clarity on how it could work, I
can try with a less intrusive PR and we can agree on the strategy before. I
will try to do it tonight.

Ste

stefanofornari:
one question nick: so far I tool the route of having a dedicated Console
class to manage the input from the console and feed Interpreter. would you
prefer instead to mix the two functionality in Interpreter?

Ste

nickl-:
It is not clear what two functionalities you are referring to. But lets keep the scope of this PR as small as possible it will just make it easier to reach a conclusion.

As I understand we are looking at implementations go the `ConsoleInterface` interface. I don't entirely agree that interpreter should implement this interface so this may still change. Instead we should have an inner class which serves as a default console and it can host the references. Interpreter would then have reference to a console class instead of all the streams in, out, err as it currently does. But this is beyond the scope of this PR... just something to keep in mind.

Stick to what we have here, the idea is to get this PR committed not continue development on the topic. If anything else is required we can look at those in subsequent PRs. Agreed? KISS is the word!

nickl-:
I dot't mind additional unit tests, as many as you like!

In summary and to help avoid confusion the required changes are:

  1 . CheckStyle fails the CI build (see travis).
  2. Try and avoid unnecessarily adding additional dependencies.

Unless it can be handled cleanly I would like to suggest NOT testing the `System.exit` and just to avoid it completely as it messes with the normal operation of unit tests (the other tests as well) which requires that a unit test complete if it is to succeed. I don't see any value in testing `System.exit` we are confident it will succeed. Lets test the other routes instead which are more likely to present problems.

3. Prevent tests from outputting to stdout or stderr.

nickl-:
You will have to rebase, just committed changes to Interpreter. I tried to hold on to these but my patches started overlapping and getting out of hand.

Shout if you need a hand.

stefanofornari:
no worries, apply your changes. I progressed in a slightly different
direction, I would probably change the PR quite a bit.

Ste

nickl-:
@stefanofornari as you wish...

I was actually hoping to get something of yours committed. Doing a huge PR is not going to make hat easy. Why don't you start with some unit tests to get a feel for the process?

stefanofornari:
hey @nickl- , I am going to try to update this pr this weekend. Do you plan to update bsh.Interpreter?

nickl-:
This was replaced by #85, right? Can we close it?

stefanofornari:
yep!

Ste

nickl-:
closed replaced by #85

PR #85, opened 2018-07-07
Delegating the display of the prompt to prompt()

stefanofornari:
Delegating the display of the prompt to prompt() so that subclasses can apply specialized logic. This may seem just a rename of a method, but it actually swaps the responsibility to display the prompt to subclasses. For example jline has its own way to manage the prompt, thus a subclass that wants to use jline as terminal, needs a way to know which prompt shall be used.

PS1: sorry, I creatng a new PR because I screwed my old branch
PS2: this does not include anything for Interrupting the interpreter (aka Ctrl+C trail); I will create another PR for it

### DIFF
--- a/src/main/java/bsh/Interpreter.java
+++ b/src/main/java/bsh/Interpreter.java
@@ -110,6 +110,7 @@ public class Interpreter
         turns it on or off.
     */
     public static final ThreadLocal<Boolean> DEBUG = ThreadLocal.withInitial(()->Boolean.FALSE);
+    private boolean EOF;
     public static boolean TRACE;
     public static boolean COMPATIBIILTY;
 
@@ -156,6 +157,9 @@ public class Interpreter
     /** The name of the file or other source that this interpreter is reading */
     String sourceFileInfo;
 
+    /** thread yield time in milliseconds */
+    private int yield_for = -1;
+
     /** by default in interactive mode System.exit() on EOF */
     private boolean exitOnEOF = true;
 
@@ -197,7 +201,7 @@ public class Interpreter
         parser = new Parser( in );
         long t1 = 0;
         if (Interpreter.DEBUG.get())
-            t1=System.currentTimeMillis();
+            t1=System.nanoTime();
         this.in = in;
         this.out = out;
         this.err = err;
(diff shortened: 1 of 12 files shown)


src/main/java/bsh/Interpreter.java
```diff
@@ -1192,6 +1192,18 @@ private String getBshPrompt()
...
+     * Interpreter calls <code>prompt()</code> when the prompt shall be displayed
+     * on console. The default implementation is that prompt is just printed to
+     * the Console/Interpreter's output stream, but subclasses with a more
+     * sophisticated prompt handling (like terminals) can hook here their logic.
+     *
+     * @param prompt the prompt to me displayed
+     */
+    protected void prompt(String prompt) {
+        print(prompt);
+    }
```
nickl-: Can we make prompt a ConsoleInteface method?

And implement this method similar to error and print:

```
if ( null != console ) {
   console.prompt(prompt);
else
   print(prompt);
}
```

This way the console implementation, ie for jline, can implement its prompt method as it sees fit.

nickl-: What I was thinking is an inner class `DefaultConsole` or something, which implements `ConsoleInteface` and holds the references to `in`, `out` and `err` with the interpreter getIn, getOut, getErr, print, println, error delegating to the innerclass instance referenced at console.

If you want to start on that this would be awesome!!! Really needed!

But if you are pressed for time just the `prompt` method as a default method on interface for now would be great and gets this PR committed.

stefanofornari:
I agree. Would you keep both print() and println()? I would vote for
print() only n the interface.

Ste

nickl-:
> keep both print() and println()? I would vote for print() only in the interface.... mmhhh... do we actually need print and error at all? or shall we drop them in favour of getOut().print[ln] and getErr().print[ln]()

We cannot change public api... that will break it for everybody. Hence why the addition of the prompt should be a default interface method (as apposed to an abstract interface method), so that it doesn't break the public interface.

We also can't remove the public methods on `Interpreter` so we will just delegate them... but we can however remove the package private fields `in`, `out` and `err`... something we wouldn't be able to do if someone went and made them `protected` members.

We don't have any choice in the matter.

stefanofornari:
omg! I am blushing... this tells me how old I am... :( got it now.

nickl-:
> omg! I am blushing... this tells me how old I am... :( got it now.

With reference to the default methods I presume... I thought you would like it =)

Closing in favour of #87

PR #86, opened 2018-07-07
Interruptable interpreter

stefanofornari:
This is another PR to enable better integration into other software, originally bshconsole. The idea is to be able to tell the interpreter the parsing operation shall be interrupted and input discarded.
The strategy is still to provide a minimal implementation in bsh.Interpreter just not to break anything. The sophisticated logic will be implemented subclassing bsh.Interpreter and overriding interrupt().
To make it useful, we need to open a few fields to subclasses.

With this little changes, together with pull #85 , bshconsole can do a fairly decent job as a console.

However, it is just a first step. The main issue I see is that interrupting the parser by closing its inputstream requires the is to be re-created. With the current design of Interpreter this is tricky. We should probably introduce an InputStream factory so that more logic can be included in the base implementation.

### DIFF
--- a/src/test/java/bsh/InterpreterTest.java
+++ b/src/test/java/bsh/InterpreterTest.java
@@ -61,7 +61,6 @@ public class InterpreterTest {
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
         Interpreter bsh = new Interpreter(in, new PrintStream(baos), System.err, false);
-        bsh.setShutdownOnExit(false);
 
         Thread t = new Thread(bsh);
         t.start();
@@ -90,7 +89,6 @@ public class InterpreterTest {
         final ByteArrayOutputStream err = new ByteArrayOutputStream();
 
         Interpreter bsh = new Interpreter(in, new PrintStream(out), new PrintStream(err), false);
-        bsh.setShutdownOnExit(false);
 
         //
         // A ParseException after interrupt() is most likely due to the


src/main/java/bsh/Interpreter.java
```diff
@@ -1158,6 +1167,27 @@ public void setErr( PrintStream err ) {
...
+     * Once interrupted the input stream and the parser become invalid and can
+     * not be used any more.
+     *
+     * Override this basic implementation for more sophisticated policy.
+     */
+    protected void interrupt() {
+        println("... interrupted ...");
+
+        try {
+            interrupted = true;
```
stefanofornari: If we had a factory method here, we could close the is and recreate the parser. Currently we just close the stream which means the parser won't work any more. Subclasses will have to override this method to recover the parser.

src/test/java/bsh/InterpreterTest.java
```diff
@@ -52,4 +55,63 @@ public void check_system_object() throws Exception {
...
+        // some more refactoring, let's defer it to the next iteration
+        //
+    }
+
+    @Test
+    public void interrupt_on_parsing() throws Exception {
+        final StringReader in = new StringReader("class A {\n"); // generate a ParseException
+        final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        final ByteArrayOutputStream err = new ByteArrayOutputStream();
+
```
stefanofornari: I hope this noisy code shows the need of some more advanced TDD tools... creating BAOS is not fun! :D

stefanofornari:
I have noticed the errors... I am checking.

src/main/java/bsh/Interpreter.java
```diff
@@ -143,11 +143,12 @@ public static void setShutdownOnExit(final boolean value) {
 
     /* --- Instance data --- */
 
-    transient Parser parser;
+    protected transient Parser parser;
+    protected transient Reader in;
+    protected transient PrintStream out;
+    protected transient PrintStream err;
+
```
nickl-: The problem with doing this...and the reason I pointed out the that it is suggested to make all public classes final (so that you cannot extend them) is because we need to be extra careful not to change any public members as they are breaking. The same goes for making members protected, we now cannot change them in the future as we have to consider subclasses in use.

The package private is a good compromise as it doesn't completely disable the ability to extend the class or inherit its members but neither does it advertise it as "supported api"

Personally I think parser should be private, it can change the reliability of the language if we allow access to it. Also the in, out and err should be referenced from console instead IMO

src/main/java/bsh/Interpreter.java
```diff
@@ -525,6 +527,12 @@ public void run()
                 if(!interactive)
                     eof = true;
             }
+            catch (TokenMgrException e) {
+                if ( DEBUG.get() )
+                    e.printStackTrace();
+                if(!interactive)
+                    eof = true;
+            }
             catch(Exception e)
```
nickl-: You were still explaining to me why we need TokenMgrException now, I forget.

src/main/java/bsh/Interpreter.java
nickl-: Can we not use eof instead? In other words, if it is interrupted it is eof.

src/test/java/bsh/InterpreterTest.java
```diff
@@ -52,4 +55,61 @@ public void check_system_object() throws Exception {
...
+        //
+        // now the interpreter should be on a blocking read; we interrupt it,
+        // give it some time to make the magic and make sure it tells us about it
+        //
+        bsh.interrupt(); Thread.sleep(250);
+
+        assertTrue(baos.toString().contains("... interrupted ..."));
+        assertTrue(!baos.toString().contains("Parser Error:"));
+
```
nickl-: So basically interrupted is used to avoid parser errors? The new implementation set EOF on close and allows readLine (previously isEOF) to wait for closing thread. This still allows errors to be reported if the interruption was not successful instead of just ignoring it to a flag.

src/test/java/bsh/InterpreterTest.java
nickl-: I am open to suggestions just needs to be motivated. I don't think there are many tests currently using this  method so shouldn't be a major issue.

src/test/java/bsh/InterpreterTest.java
```diff
@@ -52,4 +55,61 @@ public void check_system_object() throws Exception {
...
+        //
+        // A ParseException after interrupt() is most likely due to the
+        // interruption and the error should be muted
+        //
+        bsh.interrupt(); // to make it easy let's make sure we start interrupted
+
+        Thread t = new Thread(bsh);
+        t.start();
+
+        Thread.sleep(250); // give Interpreter eom time to set itself up
```
nickl-: This is how I got iTerm to drop the prompt, by trying to slow the main thread. =(

src/main/java/bsh/Interpreter.java
```diff
@@ -1254,5 +1284,14 @@ public static String getSaveClassesDir() {
...
+    public void resetParser(Reader in) {
+        try {
+            this.in.close();
+        } catch (IOException x) {
+            // nothing to do...
+        }
+        this.in = in;
+        parser = new Parser(in);
+    }
+}
```
nickl-: This should not be necessary, there is no need to reset the parser. Interrupt should interrupt the running thread not just the parser.

src/main/java/bsh/Interpreter.java
stefanofornari: I was using eof at the beginning, but it was the cause why "class A {\n ^C" printed a parser error. I realized that when you close the stream after the parser parsed "why class A {\n", Line() returns still false therefore you go on trying to parse the node and the parser throws a ParserException consuming <EOF> as a token (this is why you needed to parse the error message). Maybe it can be fixed in the grammar... I do not know.

src/test/java/bsh/InterpreterTest.java
stefanofornari: The suggestion is using junit system-rules (https://stefanbirkner.github.io/system-rules/). the major issue is that there are too few test ;), and making them easier to write will help.

src/main/java/bsh/Interpreter.java
stefanofornari: well, I may agree in principle (even if only up to a certain point), but to me this is mostly upfront thinking. do we really have a problem of tons of people creating subclasses of Interpreter? I would really love so, but may be it won't be the case for the next few months. In any case, please note that they are not private now. They have package visibility. This does not give you the benefits you are wishing, and makes subclassing odd (why foo project should have a bsh package to extend Interpreter?)

src/main/java/bsh/Interpreter.java
stefanofornari: Sorry, it's a leftover. I will remove it.

src/test/java/bsh/InterpreterTest.java
stefanofornari: interrupted will be to tell interpreter at any level that the current input or processing should be interrupted. see my first comment on why we can not use eof (unless now Line() has a different behavior). Let me know if not clear.

src/test/java/bsh/InterpreterTest.java
stefanofornari: That I think is ok for th tests only. What I have done in BshConsoleInterpeter is: 

in startConsole():

synchronized (this) {
            try {
                wait(2000);
            } catch (InterruptedException x) {
                // nop
            }
        }

in prompt():
synchronized (this) {
            notifyAll();
        }

This can probably only achieved outside Interpreter...

src/main/java/bsh/Interpreter.java
stefanofornari: This is where we have different views. I really think we should not stop REPL and recreate everything. I have seen your other comment about closing and recreating the parser not costing to much, but it simply does not sound correct. It may be a workaround to achieve our goal, but I believe there is a better way. Interpreter shall have the responsibility to get the input parse it and run the parsed code keeping the state of all this until you have finished your session.
The Parser object is what is used to parse a portion of the input and its lifecycle is shorter than Interpreter. We want to stop this last part, not everything.

said that, I was re-reading something about Thread.interrupt, and we may have the solution that works for both: 
>>
The Interrupt Status Flag

The interrupt mechanism is implemented using an internal flag known as the interrupt status. Invoking Thread.interrupt sets this flag. When a thread checks for an interrupt by invoking the static method Thread.interrupted, interrupt status is cleared. The non-static isInterrupted method, which is used by one thread to query the interrupt status of another, does not change the interrupt status flag.
<<
If we use Thread.interrupt to tell everything to stop what they are doing and then we clear the status without exiting from REPL, we would be good. The only caveat will still be we need to create a new parser, which to me seems perfectly fine.

src/test/java/bsh/InterpreterTest.java
nickl-: Yes I understand why you are using `boolean interrupted` but all it is is a flag, it does not do anything nor does it guarantee anything. `Thread.interrupted` guarantees that the Thread will end soon. `EOF` means we have reached EOF and can abort the repl, they are both truthy and falsey but they already have tied to functionality. So why use something else which yet has no functionality other than true or false?

nickl-:
> The interrupt mechanism is implemented using an internal flag known as the interrupt status. Invoking Thread.interrupt sets this flag. When a thread checks for an interrupt by invoking the static method Thread.interrupted, interrupt status is cleared. The non-static isInterrupted method, which is used by one thread to query the interrupt status of another, does not change the interrupt status flag.


> If we use Thread.interrupt to tell everything to stop what they are doing and then we clear the status without exiting from REPL, we would be good. The only caveat will still be we need to create a new parser, which to me seems perfectly fine.

I don't think you understood it correctly. The thread methods `stop` `suspend` , `destroy` and `resume` have all been deprecated and replaced by `interrupt`. It does not immediately kill the thread but it sets the static field `interrupted` by which the thread will at it's earliest convenience finish what it was doing and die.

So you can't both interrupt the thread and keep it running. The loop interrupt worked without thread interrupt and the additions of `Thread.interupted` to the loops, all this does is give the thread a good excuse to finish sooner, it is going to die regardless. We are working on the new interpreter while it does its thing.

nickl-:
**Closed: won't implement**

PR #87, opened 2018-07-08
New console interface

stefanofornari:
This PR extends the concept introduced in pull #85, extending the ConsoleInterface interface and using it in INterpreter as opposed to implementing it. It required some additional work and must be reviewed in two areas:

1. serialization/deseraialization
2. ScriptEngine implementation

The PR introduces a SimpleConsole that can hold any stream and a StandardConsole that holds system streams only. This may be good for tests only... I wanted to use it in Interpreter too when no stream is specified in constructors, but I ran into unexpected serialization issues.

I have three main open issues:

1. setOut, setErr: they seem to be used by the ScriptEngine JSR, so I added it in the interface as opposed to methods in Interpreter. but I am not sure I like it. the interface is more cluttered in this way.
2. I had to @Ignore for now test_evakl_writer_unicode in TestBshScriptEngine; before working on it I would like to get feedbacks on the rest of the work
3. There is a setConsole() method that rightly asks if when somebody sets the console we should reinit the parser.... a legitimate question...

I may add comments inside the code as well.

### DIFF
--- a/src/main/java/bsh/Interpreter.java
+++ b/src/main/java/bsh/Interpreter.java
@@ -526,12 +526,6 @@ public class Interpreter
                 if(!interactive)
                     eof = true;
             }
-            catch (TokenMgrException e) {
-                if ( DEBUG.get() )
-                    e.printStackTrace();
-                if(!interactive)
-                    eof = true;
-            }
             catch(Exception e)
             {
                 error("Unknown error: " + e);


src/main/java/bsh/Interpreter.java
```diff
@@ -291,10 +292,10 @@ public Interpreter()
     public void setConsole( ConsoleInterface console ) {
         this.console = console;
         setu( "bsh.console", console );
-        // redundant with constructor
-        setOut( console.getOut() );
-        setErr( console.getErr() );
-        // need to set the input stream - reinit the parser?
```
stefanofornari: I removed the comment, but maybe we should keep it? the question is valid, we probably should, otherwise there is no point in setting a new Console.

src/main/java/bsh/Interpreter.java
```diff
@@ -460,7 +461,7 @@ public void run()
 
                 eof = isEOF();
 
-                if( get_jjtree().nodeArity() > 0 )  // number of child nodes
+                if( !Thread.interrupted() && get_jjtree().nodeArity() > 0 )  // number of child nodes
```
stefanofornari: this does not really have to do with this pr...

src/main/java/bsh/Interpreter.java
```diff
@@ -491,13 +492,13 @@ public void run()
             }
             catch(ParseException e)
             {
-                if ( !interactive || !e.getMessage().contains("<EOF>") )
+                if ( !interactive && !Thread.interrupted() )
```
stefanofornari: neither this...

src/main/java/bsh/engine/BshScriptEngine.java
```diff
@@ -362,14 +360,14 @@ public Object invokeFunction(String name, Object... args) throws ScriptException
         Interpreter.
     */
 
-    class WriterOutputStream extends OutputStream {
+    static public class WriterOutputStream extends OutputStream {
```
stefanofornari: I was trying to use this in my test, but I could not figure it out... I will come back on this if the rest is ok.

src/main/java/bsh/util/AWTConsole.java
```diff
@@ -350,4 +349,24 @@ public String toString() {
     // unused
     public void keyTyped(KeyEvent e) { }
     public void keyReleased(KeyEvent e) { }
+
```
stefanofornari: here I haven't taken out the streams yet. let's finalize Interpreter first, than we can come back to this one.

src/main/java/bsh/util/JConsole.java
```diff
@@ -820,6 +819,26 @@ public void setWaitFeedback( boolean on ) {
 
     private int textLength() { return text.getDocument().getLength(); }
 
+    @Override
```
stefanofornari: here I haven't taken out the streams yet. let's finalize Interpreter first, than we can come back to this one.

nickl-:
> setOut, setErr: they seem to be used by the ScriptEngine JSR, so I added it in the interface as opposed to methods in Interpreter. but I am not sure I like it. the interface is more cluttered in this way.

Ah yes... I was wondering about these methods. We are going to have to get creative with these. Remember we cannot break the interface!!!! If I have a class that implements the interface it should shill compile after your modifications.

nickl-:
Well done! Great work!

Shout if you have any questions otherwise I think you have enough to go on with, right?

src/main/java/bsh/console/StandardConsole.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public void prompt(String prompt) {
+        print(prompt);
+    }
+
+    @Override
+    public void setIn(Reader in) {
+        throw new IllegalArgumentException("System.in can not be overriden");
+    }
```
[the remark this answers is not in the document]
nickl-: hmm these might actually make good default methods... see ConsoleInteface comments

src/main/java/bsh/ConsoleInterface.java
```diff
@@ -37,11 +37,21 @@
...
+
+    public void prompt(String prompt);
+
+    //
+    // shall we deprecate these?
+    //
+    public void setIn (Reader       in);
+    public void setOut(PrintStream out);
+    public void setErr(PrintStream err);
 }
```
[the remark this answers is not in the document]
nickl-: These could also make suitable default methods... on second thought

```java
default void setIn(Reader in) {
       throw new IllegalArgumentException("Not implemented"); // is there not a NotImplemented or Unimplemented Exception?
}
```

src/main/java/bsh/PreparsedScript.java
```diff
@@ -1,12 +1,11 @@
 package bsh;
```
nickl-: I have a feeling if this requires any change outside of `Interpreter` then it is a sign that we are doing something wrong.

stefanofornari:
ouch! thanks god you are not by boss!!! :)
I will go through the comments and fix them over the week. thanks a lot for your feedbacks. I pretty much agree on everything.

stefanofornari:
@nickl- m one question though: you were suggesting to add prompt() to ConsoleInterface, but this will break compatibility and other classes. We can create a sub-interface and use instanceof in Interpreter, but it is not very nice. Suggestions?

stefanofornari:
Hi @nickl- , I've been trying with a subinterface, but I could not find a way that works properly without breaking compatibility. setOut()/setErr()  cannot be implemented without accessing directly the streams... nor you can use a ConsoleInterface because there is not a contract for setting the streams.
At this point, I would give up and I would request to merge pull #85.

nickl-:
> Hi @nickl- , I've been trying with a subinterface, but I could not find a way that works properly without breaking compatibility. setOut()/setErr() cannot be implemented without accessing directly the streams... nor you can use a ConsoleInterface because there is not a contract for setting the streams.


I gave you 2 options that will work, default method or subinterface. Why is this not working? 

Please push what you have so that I can have a look.

nickl-:
> @nickl- m one question though: you were suggesting to add prompt() to ConsoleInterface, but this will break compatibility and other classes. We can create a sub-interface and use instanceof in Interpreter, but it is not very nice. Suggestions?

I said [default interface method](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html). I even gave you the code snippet  to use.

nickl-:
Repost:

Try

```java
default void prompt(String prompt) {
    print(prompt);
}
```
or something like that

stefanofornari:
> Try
> 
> default void prompt(String prompt) {
>     print(prompt);
> }

I am really confused. the point is not the visibility of the method. If we add a method to ConsoleInterface, any current implementation of that interface will break. So if you do not want to make incompatible changes, there is no way to use ConsoleInterface.

nickl-:
Closed: in favour of #88

PR #88, opened 2018-07-11
New console interface - first step

stefanofornari:
Hi @nickl- ,

I am sorry, I screwed again with the PR :(, so I am creating a new one. Scope is restricted to the management of the prompt.

As per setOut/setErr/setIn, I am still doubtful it would work: if I pass in an old ConsoleInterface to Interpreter, calling setOut will delegate to the console's setOut (because console is != null), which will not give the expected behaviour. 

Sorry again for the overhead of the new PR.... (the good news is that I am learning... :) )


src/main/java/bsh/ConsoleInterface.java
```diff
@@ -43,5 +43,9 @@
     public void println( Object o );
     public void print( Object o );
     public void error( Object o );
+
+    default public void prompt( String prompt ) {
```
nickl-: From the [docs: default methods](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html)
All method declarations in an interface, including default methods, are implicitly public, so you can omit the public modifier.

nickl-:
> I am sorry, I missed up again with the PR :(, so I am creating a new one.

No worries, all gewd! What is up with those repeated commits with the same message?

If you want to change anything from the previous commit, it is possible to change any past commit but the previous one is very simple, just add the new changes you want and then commit with `--amend` flag. So instead of a new commit it amends the previous one.

example:

```shell
$ git commit -m"Stuff and stuff"
$ vi somefile.changed
$ git add somefile.changed
$ git commit --amend
```

By default it will open your editor with the commit message, which you can change, and the change log. Just save the commit and it will be changed. If you already pushed that commit to your GitHub repository then just add `--force` to the push command to update the remote commit reference.

```shell
$ git push origin branch-name --force
```

> Scope is restricted to the management of the prompt.

Ok fine, but we can go ahead and add the default console implementation inner class so long.

> As per setOut/setErr/setIn, I am still doubtful it would work: if I pass in an old ConsoleInterface to Interpreter, calling setOut will delegate to the console's setOut (because console is != null), which will not give the expected behaviour.

Ok so lets think what the expected behaviour will be?  Currently it would change the class instance variables `in`, `out` and `err`.  So we can leave them for now... I have a plan... lets create the default Console implementation and delegate the print, println and error methods to console. Taking it one step at a time.

 
> Sorry again for the overhead of the new PR.... (the good news is that I am learning... :) )

Really no sweats man! Submit as many PRs as you like. Glad you had more joy with the default method this time around =)

Looking good! =D

nickl-:
How about this for an Idea:

We create a new interface, lets call in `ConsoleAssignable` which extends our `ConsoleInterface`.

```java
public interface ConsoleAssignable extends ConsoleInterface {
    void setIn( Reader in );
    void setOut( PrintStream out );
    void setErr( PrintStream err );
    void setConsole( ConsoleInterface console );
}
```

We change the type of the `console` variable in `Interpreter` to `ConsoleAssignable` and basically just move all the functionality from `Interpreter` to the inner class `Interpreter.Console` which implements the `ConsoleAssignable` interface.

Something like this:

```java
    public static class Console implements ConsoleAssignable {
        public static String systemLineSeparator = "\n";
        private transient Reader in;
        private transient PrintStream out;
        private transient PrintStream err;
        private ConsoleInterface console;

        public Console(ConsoleInterface console) {
            this.console = console;
            this.in = console.getIn();
            this.out = console.getOut();
            this.err = console.getErr();
        }
        
        public Console(Reader in, PrintStream out, PrintStream err) {
            this.console = null;
            this.in = in;
            this.out = out;
            this.err = err;
        }

        @Override
        public Reader getIn() {
            return in;
        }

        @Override
        public PrintStream getOut() {
            return out;
        }

        @Override
        public PrintStream getErr() {
            return err;
        }

        @Override
        public void println(Object o) {
            if ( null != console )
                console.println(o);
            else
                this.print( o + systemLineSeparator );
        }

        @Override
        public void print(Object o) {
            if ( null != console )
                console.print(o);
            else if ( null != out ) {
                out.print(o);
                out.flush();
            }
        }

        @Override
        public void error(Object o) {
            if ( null != console )
                console.error( "// Error: " + o + systemLineSeparator );
            else if ( null != out ) {
                this.println( "// Error: " + o );
            }
        }

        @Override
        public void prompt(Object o) {
            if ( null != console )
                console.prompt(o);
            else
                this.print(o);
        }

        @Override
        public void setIn(Reader in) {
            this.in = in;
        }

        @Override
        public void setOut(PrintStream out) {
            this.out = out;
        }

        @Override
        public void setErr(PrintStream err) {
            this.err = err;
        }

        @Override
        public void setConsole(ConsoleInterface console) {
            this.console = console;
            this.setOut( console.getOut() );
            this.setErr( console.getErr() );
            this.setIn( console.getIn() );
        }
        
    }
```

With 2 constructors by which we can instantiate our `Interpreter` local `console` reference with either a `ConsoleInterface` or the `in`, `out` and `err` streams whichever is available and everything should continue to work as before.

This should do the trick... what do you think? Did I miss anything?

stefanofornari:
> check the redundant use of public in the default method, public can actually be removed from the abstract methods (print, println, print) in the interface too.

done.

> Please add Interpreter.Console inner class.

I haven't done it yet, because until we understand how to manage the setXXX(), I am not sure it is  worth.

You can reference to https://github.com/stefanofornari/beanshell/commit/2e2d645fb3dddc61c8c4cb91fede0002413c25bb for a tentative implementation that went in the direction you mentioned. Where I am stuck is that Interpreter.setOut() in combination with setConsole().
If we make setOut() delegate to console.setOut() we force the console to be assignable, which would break compatibility. If we do not delegate, then we can not use an inner default implementation, plus if a client calls setConsole with an assignable console the result will not be as expected... do you see what I mean?

PS: thanks for the git hints.

src/main/java/bsh/ConsoleInterface.java
```diff
@@ -37,11 +37,15 @@
...
+    PrintStream getOut();
+    PrintStream getErr();
+    void println( Object o );
+    void print( Object o );
+    void error( Object o );
+
+    default void prompt( String prompt ) {
+        print(prompt);
+    }
 }
```
nickl-: Very nice! That looks so much better. Good job!

src/main/java/bsh/Interpreter.java
```diff
@@ -468,7 +468,7 @@ public void run()
             try
             {
                 if ( interactive )
-                    print( getBshPrompt() );
+                    prompt(getBshPrompt());
```
nickl-: This should be able to call the default method without changing it.

src/main/java/bsh/Interpreter.java
```diff
@@ -825,6 +827,15 @@ public final void print( Object o )
...
+    @Override
+    public final void prompt(String prompt) {
+        if (console != null) {
+            console.prompt(prompt);
+        } else {
+            print(prompt);
+        }
+    }
+
     // End ConsoleInterface
```
nickl-: Lets not implement this here, would you agree?

We can use the default method for now and avoid creating another public api method on Interpreter which we can't delete. Once we implement the default console class we can elaborate on that. For now the inherited method from `ConsoleInterface` which delegates to print will suffice.

So please remove this method implementation.

nickl-:
By "Console class is not required" I mean to say that it is not required to commit this PR... the Console class itself is very much needed. Just to make sure =)

nickl-:
@stefanofornari did you manage to remove the prompt method from Interpreter? Is everything still working then?

I am ready to complete this when you are.

stefanofornari:
@nickl- done.

nickl-:
Already merged, PR can be closed. see beanshell/beanshell@4d21e02

Also implemented the Interpreter changes, extended interface, inner class `Interpreter.Console`, non breaking fix. see beanshell/beanshell@9b4d2d7

@stefanofornari  Note: `setConsole` now also reinitialises the parser, perhaps this would work for your Ctrl+C implementation. Let me know....

Closed: PR merged

Issue #418, opened 2007-01-09
Cannot anonymous subclass with arguments


sourceforge-issue-exporter:
I have prepared a simple base class that takes a few arguments in the constructor, like this:

```java
public class TargetDescription {

    private String description;
    private String category;

    public TargetDescription(String description, String category) {
        this.description = description;
        this.category = category;
    }

    public void apply() {
    }
}
```

I have then tried to subclass this as follows:

```java
TargetDescription a = new TargetDescription("a", "b") {
    public void apply() {
        System.err.println("Hello");
    }
};
```
but when I do I get error message like the following:

```
Target exception: Sourced file: testt.bsh : Typed variable declaration : Error constructing inner class instance: bsh.ReflectError: Can't find constructor: global$1(String, String) in class: global$1 : at Line: 1 : in file: testt.bsh : new TargetDescription ( "a" , "b" ) { 
```

Reported by: *anonymous

Original Ticket: "beanshell/bugs/280":https://sourceforge.net/p/beanshell/bugs/280

PR #488, opened 2018-08-02
Add fix and test for subclass with arguments (Sourceforge issue 104075)

opeongo:
Please consider this fix for an issue involving subclassing with arguments.

### DIFF
--- a/src/main/java/bsh/BSHAllocationExpression.java
+++ b/src/main/java/bsh/BSHAllocationExpression.java
@@ -28,6 +28,7 @@
 
 package bsh;
 
+import java.io.StringReader;
 import java.lang.reflect.Array;
 import java.lang.reflect.InvocationTargetException;
 
@@ -173,10 +174,68 @@ class BSHAllocationExpression extends SimpleNode
 	{
 		//throw new InterpreterError("constructWithClassBody unimplemented");
 
-		String name = callstack.top().getName() + "$" + (++innerClassCount);
+		String inner = "anon" + (++innerClassCount);
+		String outer = callstack.top().getName();
+		String name =  outer + "$" + inner;
 		Modifiers modifiers = new Modifiers();
 		modifiers.addModifier( Modifiers.CLASS, "public" );
 		Class clas;
+
+		// If the class was created with arguments then add a constructor that simply invokes super
+		// This fails if some args are null, due to subsequent Bsh stuff
+		if (args.length > 0) {
+		   String formal = "";
+		   String argnames = "";
+		   for (int i=0; i<args.length; i++) {
+		      if (i>0) {
+			 formal += ", ";
+			 argnames += ", ";
+		      }
+
+		      String argname = String.valueOf((char) (97+i));
+		      argnames += argname;
+
+		      String className = null;
+		      String arrayDim = "";
+		      if (args[i] instanceof Primitive) {
+			 Object arg = ((Primitive)args[i]).getValue();
+			 if (arg instanceof Boolean)
+			    className = "boolean";
+			 else if (arg instanceof Byte)
+			    className = "byte";
+			 else if (arg instanceof Character)
+			    className = "char";
+			 else if (arg instanceof Double)
+			    className = "double";
+			 else if (arg instanceof Float)
+			    className = "Float";
+			 else if (arg instanceof Integer)
+			    className = "int";
+			 else if (arg instanceof Long)
+			    className = "long";
+			 else if (arg instanceof Short)
+			    className = "short";
+			 else
+			    className = "void";
+		      } else {
+			 className= args[i].getClass().getName();
+			 while(className.startsWith("[")) {
+			    arrayDim += "[]";
+			    className = className.substring(1);
+			 }
+			 if (className.matches("L.*;")) 
+			    className = className.substring(1,className.length()-1);
+		      }
+		      formal += className+arrayDim+" "+argname;
+		   }
+		   String command = inner+"("+formal+") {super("+argnames+");}";
+		Parser p = new Parser(new StringReader(command));
+		p.Line();
+		BSHMethodDeclaration node = (BSHMethodDeclaration) p.popNode();
+		node.eval(new CallStack(), new Interpreter());
+		block.jjtAddChild(node, block.jjtGetNumChildren());
+		}
+
 		try {
 			clas = ClassGenerator.getClassGenerator() .generateClass( 
 				name, modifiers, null/*interfaces*/, type/*superClass*/,
--- a/src/test/java/bsh/SourceForgeIssuesTest.java
+++ b/src/test/java/bsh/SourceForgeIssuesTest.java
@@ -170,6 +170,33 @@ public class SourceForgeIssuesTest {
     }
 
 
+   /**
+    * <a href="http://sourceforge.net/tracker/?func=detail&aid=1631576&group_id=4075&atid=104075">Sourceforge issue "Cannot anonymous subclass with arguments" - ID: 1631576</a>
+    */
+   @Test
+   public void sourceforge_issue_104075() throws Exception {
+      final String CODE_104075 =
+	 "public class SubclassWithArgs {\n"+
+	 "   protected String arg1;\n"+
+	 "   public  SubclassWithArgs(String a) {\n"+
+	 "      arg1 = a;\n"+
+	 "   }\n"+
+   
+	 "   public String b() {\n"+
+	 "      return \"Goodbye\";\n"+
+	 "   }\n"+
+	 "}\n"+
+	 
+	 "scwa = new SubclassWithArgs(\"world\") {\n"+
+	 "   public String b() {\n"+
+	 "       return \"Hello \"+arg1;\n"+
+	 "   }\n"+
+	 "};\n"+
+	 // "System.err.println(scwa.b());\n"+
+	 "return scwa.b();\n";
+      assertEquals("Hello world", eval(CODE_104075));
+   }
+
     public static class A {
 
         public static String staticMethod() {


nickl-:
Wow! Nice one... well spotted.

I did a few tests as well to see if I can reproduce this. I chose 3 different parent classes to test with to see if the presence of a default parent constructor has any effect or not. I also wanted to see what `this.getClass()` will produce..

## Parent classes
```java
// parent class with default constructor
class DefaultConstructor {
    cls = this.getClass();
    DefaultConstructor() { }
}

// parent class without default constructor
class NoDefaultConstructor {
    cls = this.getClass();
    NoDefaultConstructor(String  a) { }
}

// parent class with default and other constructor
class OvrLoadedConstructor {
    cls = this.getClass();
    OvrLoadedConstructor() { }
    OvrLoadedConstructor(a)  { }
}
```
The following results were as expected.
```java
c = new DefaultConstructor();
assertThat(c.cls, equalTo(DefaultConstructor.class));
c = new NoDefaultConstructor("foo");
assertThat(c.cls, equalTo(NoDefaultConstructor.class));
c = new OvrLoadedConstructor();
assertThat(c.cls, equalTo(OvrLoadedConstructor.class));
c = new OvrLoadedConstructor("foo");
assertThat(c.cls, equalTo(OvrLoadedConstructor.class));
```

## Sub-classes
I think your wording is slightly off, not a biggy but for our project in particular we need to label these things consistently. Yes it's true and I agree 100% sub classes in the general sense are classes that inherit from a super class (or multiple super classes in some cases) so these are all sub classes but in the java domain sub classes are distinguishable from other types of classes and those specifically being the classes that extend from a super class, like this one. 
```java
class SubDefaultConstructor extends NoDefaultConstructor { }
```
This guy is of course broken and we cannot get an instance with it. If we try with no arguments it cannot find the parent default constructor.
```java
sub = new SubDefaultConstructor();
```
#### error message
```
NoDefaultConstructor: method <init>()V not found
```
And if we try with an argument it will complain that it cannot find the sub class constructor.
```java
sub = new SubDefaultConstructor("foo");
```
#### error message
```
Can't find constructor: SubDefaultConstructor(String) in class: SubDefaultConstructor
```

This is the same in java so no need for concern. If we want to look at some working examples lets have one that overrides and one tho overloads the parent constructor. I also wanted to see what they report as their super class. 

```java
class SubOverrideSuper extends NoDefaultConstructor {
    spr = this.getClass().getSuperclass();
    SubOverrideSuper(a) { }
}
class SubOverloadSuper extends NoDefaultConstructor {
    spr = this.getClass().getSuperclass();
    SubOverloadSuper() {  }
}
```
The results show that `this.getClass()` return the sub class and `this.getClass().getSuperclass()` reports their common parent. Everything seems to be working as expected so far.
```java
sub = new SubOverrideSuper("foo");
assertThat(sub.cls, equalTo(SubOverrideSuper.class));
assertThat(sub.spr, equalTo(NoDefaultConstructor.class));
sub = new SubOverloadSuper();
assertThat(sub.cls, equalTo(SubOverloadSuper.class));
assertThat(sub.spr, equalTo(NoDefaultConstructor.class));
```

## anonymous classes
The problem @opeongo discovered relates to anonymous classes. They are a special kind of sub classes which extend a super class and the same inheritance rules apply but with one difference they have no name (class).

They do still have unique names reported by `this.getClass()` but there is no named class we can load for them. Lets also look at what they report their super classes to be shall we.

The default constructor anonymous class works.
```java
anon = new DefaultConstructor() {
    spr = this.getClass().getSuperclass();
};
assertThat(anon.cls.getName(), startsWith("global"));
assertThat(anon.spr, equalTo(DefaultConstructor.class));
```

Same goes for the no arguments constructor from the overloaded class.
```java
anon = new OvrLoadedConstructor() {
    spr = this.getClass().getSuperclass();
};
assertThat(anon.cls.getName(), startsWith("global"));
assertThat(anon.spr, equalTo(OvrLoadedConstructor.class));
```

But as soon as we supply constructor parameters it fails. 
```java
anon = new NoDefaultConstructor("foobar") { };
anon = new OvrLoadedConstructor("foobar") { };
```
Both produce the following error.
#### error message
```
Can't find constructor: global$2(String) in class: global$2
```

Same exception as with our failing subclass, these classes only have default constructors and cannot be instantiated with arguments. Unfortunately for us this is possible with java so we indeed have a problem which needs to be resolved.

nickl-:
I will start a thorough review of this patch shortly just a heads up regarding the unit tests. I've been trying to break away from generic test names or referencing issue numbers as unit tests. It may make sense at time of implementation but looses it's intent and purpose very soon there after. The only reason why I still haven't managed to rid them all is because I can't be sure under what categories these tests actually belong.

So please find a name more appropriate than sourceforge issue xxx or create a new unit test. You are also welcome to add this test script with the tests from the previous post. 

I'm not saying you have to do tests like this, if you prefer junit tests it is all good, call it a thank you for your contribution gift =)  

[anonymous_sub_classes.bsh.txt](https://github.com/beanshell/beanshell/files/2255417/anonymous_sub_classes.bsh.txt)

Remove the .txt extension and save it under `src/test/resources/test-scripts/` and it will automatically be included with the other test scripts. If you want to single out this test which also produces more verbose errors you can use the following maven command.

```
mvn test -Dtest=BshScriptTestCase -Dscript=anonymous_sub_classes.bsh
```

nickl-:
Fixes #418

opeongo:
Ok, so my patch was a quick hack to show the problem and where the fix would have to be made.  Not elegant and not complete.  Thaks for having a look at it and for the thorough review.

Great idea to reuse the existing code to find the most specific constructor in the super class.  This is more robust and will avoid the problem I had with null parameter values.  I may need to refactor a method in Reflect.java for this.

We need to have evaluated arguments to figure out the constructor signiture, so the addition of the constructor with parameters has to be done during the eval phase, and cannot happen when the AST is initially being built.  I think I have the correct location in the code to make the fix.

The constructor with arguments is a BSHMethodDeclaration node, and this has to be included in the BSHAllocationExpression block.  What my code is doing is building a set of nodes that implement a BSHMethodDeclaration, and then adds this as a child to the block.  In order to avoid using the parser I can build this 'by hand' but it is a bit lengthy.  

The BSHMethodDeclaration requires a BSHAmbiguousName, BSHReturnType, BSHFormalParameters and BSHBlock and Modifiers,  
BSHReturnType requires a BSHType.
BSHType requires a BSHAmbiguousName. 
BSHFormalParameters requires an array of BSHFormalParameter nodes.  
BSHFormalParameter requires a BSHType.
BSHBlock requires (I'm not sure, will have to look harder)
BFormatParameter requires a BSHType.

Here is what it is starting to look like:
```java
		   // Create the constructor with args
		   BSHMethodDeclaration bmd = new BSHMethodDeclaration(0);

		   // Set the name to be the class name
		   bmd.name = inner;
		   
		   // set the return type
		   BSHReturnType brt = new BSHReturnType(0);
		   BSHType bt = new BSHType(0);
		   BSHAmbiguousName ban = new BSHAmbiguousName(0);
		   ban.text = inner;
		   bt.jjtAddChild(ban, 0);
		   brt.jjtAddChild(bt, 0);
		   bmd.returnTypeNode = brt;
		   
		   // set the formal parameters
		   BSHFormalParameters bfp = new BSHFormalParameters();
```
This is all done automatically in the parser by jjtree, so I don't think that there are any bsh support routines that I can reuse.  If I do implement by hand-building the tree, it will be brittle and suceptable to breakage if the parser is changed.

I realize that it seems a bit wierd to make a string and parse it when we have all of the parts, but there is quite a bit to do to make the tree correctly, and no helper routines.  Using the parser is just a couple of lines and generates a correct tree.

Let me know what you prefer.

opeongo:
After stepping away and thinking about it, the correect action is to add all of the non-default constructors to the block when the node is closed.  This does not require evaluation of the arguments.  

Also, it should be relatively simple to code up a NodeUtils class to help construct the constructor nodes, so no need to make a string and call the parser.  I'll give this a try after the weekend.

nickl-:
The code generation/parsing trix is really a novel idea, had me intrigued for a while. I am sure there is application for it in the world. It does not take a leap of imagination to conclude that the machine will eventually write its own code... and its language of choice is BeanShell =) 

Boiling down to brass tax though BeanShell is an interpreter which clearly sets code generation outside of scope. I could imagine it as a BeanShell side project perhaps but none the less interpreter would never require it as a dependency. IMHO

To produce a functional constructor for the child requires three parts, a method declaration, the declared methods invocation (method block evaluation with arguments) and a method invocation to `super` as a function call of the method block.

### method declaration
BeanShell compiles method declarations as their implementations (`BshMethod`) so it would be possible to produce an equivalent method without needing an AST definition, we currently wrap reflected java  methods in this fashion so there is precedence.

Aside from method name, the equivalent parent constructor holds everything necessary to produce a functional constructor declaration for the child. Finding the parent constructor is by means of `Reflect.findMostSpecificConstructor` and works as advertised without the need for any modification for application to these requirements.

`DelayedEvalBshMethod` is a wrapper to `BshMethod` which collects all the necessary parts to successfully produce a method at a later stage, a collection of which is used to produce constructors and methods during class generation. With the addition of an overloaded `DelayedEvalBshMethod` constructor to take the super constructor as argument I was able to produce the method declaration.

BeanShell makes no effort to extract anything from method blocks at this point, they are simply referenced by the extracted method for evaluation at invocation.

### declared method invocation
Methods are invoked with argument values no surprises there, and returns a value based on its return type.

For java wrapped methods the arguments are passed to the invoke method of the java reflected method with results returned.

For artificial methods this entails the evaluation of the referenced `Block` node (the method body) after creation of a scoped method namespace and the method argument values set as local variables to this namespace returning the result of the evaluation.
 
Constructor methods are slightly different though even though BeanShell treats constructors like methods, java reflect constructors are not invocable per se (the use `newInstace`) but `MethodHandles` on the other hand does treat constructors similar to methods but anyway. The main difference is that the 1st line of a constructor if it is a method invocation to `this` or `super` gets executed as part of the instantiation phase for object construction. Beanshell constructor type methods invoked simply ignores these calls see `BSHMethodInvocation.eval`.

Constructors are invoked as a function of an artificial class' `<init>` block or `{ }` no arguments block. BeanShell emulates instantiation by identifying the appropriate declared constructor based on supplied arguments. The 1st node of an artificial constructor`s method block gets inspected, if it is a method invocation to `super` or `this` it gets processed as a `ConstructorArgs` instance via the constructor switch. Now to tell you the truth I have not quite wrapped my head around this piece of byte code voodoo yet so this is probably high time to do so.

Unfortunately this whole process is not done as a function of the method which should probably be refactored to extract a `BshMethod.invokeInitialize`, or such, method to allow us to override this functionality with our custom constructor method.

### super invocation
As things currently stand artificial constructors need to have the super invocation identifiable as the 1st node of the method block.
The following hierarchy seems to suffice:

```
Block
    Assignment
        PrimaryExpression
            MethodInvocation
AmbiguousName        Arguments
text = "super"       args = Object[] args
```

With a small patch to `BSHArguments` to side step evaluation in the presence of an `args` instance variable it is possible to assign this node the value of the args we already have successfully producing the required method invocation.

With my initial hacks I managed to produce an instance of the anonymous class. Hooray!!! But at closer inspection it now has itself as super class so something is still not peachy. Obviously the incorrect constructor is initialised the why still needs to be discovered.

I'll make this hack available in a new branch so that we can work together on finding a better solution.

Glad we're on the same page! =)

nickl-:
Ok so turns out it was actually working, in my sleep deprived state I inadvertently cached `BSHArgument.args`, which wasn't my intention, and it was the cached print arguments that lead me to believe they were the same.

see development branch anonymous-subclasses

Using a scratch pad unit test which can be run with 

```
mvn test -Djacoco.skip  -Dtest=BshScriptTestCase -Dscript=scratch.bsh
```

Should produce the following output

```
----------------------------------------
with args: foobar bar baz
super class: class OverLoadedConstructor
this class: class global$anon1
----------------------------------------
without args: foo bar baz
super class: class OverLoadedConstructor
this class: class global$anon2
----------------------------------------
```

I agree with you that counterfeiting the AST is not the best plan, will break if the grammar changes etc, and if we can avoid that it would be better. If we refactor the following code, which is also dependent on the specific AST structure (maybe this can also be avoided),  and make it a function of the constructor then we can override that functionality to simply return the arguments instead.

https://github.com/beanshell/beanshell/blob/7800adb099fff0f1a838ccae6783f0cbdb159375/src/main/java/bsh/This.java#L531-L547

I should probably look at understanding constructor switch still. =/

nickl-:
## Constructor Switch

Basically for every constructor we generate on a virtual class we generate a switch statement to every possible this and super constructors you could possibly need. A lookup to `This.getConstructorArgs` determines which constructor you are calling with specific args as per the declared method invocation. This gets returned as a `ConstructorArgs` instance with the switch selector and delegates to return the primitive form of each argument where needed. Each possible constructor is ready to collect its arguments as per its parameter types.

nickl-:
That should do the trick see beanshell/beanshell@b044364

Now requires 0 artificial AST
Also improved the structure tight initialisation call inference
Requires no patch to Arguments node anymore
Uses a zero child method block

Kudos to @opeongo for reminding us and for showing the way, we would not have spotted this without you! On behalf of BeanShell past our sincerest apologies for the long delay (11 years) to finally get this done.

Looking forward to more PRs from you.

**Closed: fixed**

opeongo:
Hi Nick,

That looks like the correct solution to the problem.  You have a much greater comprehension of the code generation aspects than I.  My approach was papering over the problem, but you fixed it.  Thanks so much,

Tom

nickl-:
@opeongo I've been at it for a while now especially through the fork merging phase I had to quickly find out what is being replaced and what was added and needs to stay. Also doing Interfaces and Enums was an eye opening experience =)

I knew this was going to be involved so I just ran with it wanting to get you the fix you've been waiting for for so long. Knowing exactly what needed to be done of course made life much easier.

Glad you are satisfied!

Ty you for the reply! Shout if you spot anything dodgy, it still does the oddest things every now and again which I just can't put my finger on. Someone with your fault assessment skills would come in handy.

PR #495, opened 2018-08-10
Fix the situation where a \\ is at the end of a string or character literal

opeongo:
The problem was that if there was a string that ended in \\\\ (for example "abcd\\\\" or even just '\\\\') the token manager would make a mistake and swallow the final \\" (or \\') as being part of the literal, and then continue to gooble characters until the next starting " (or '), or until the end of the line.  Anyway, bad things happened.

Basically this patch is just a revert to the CHARACTER_LITERAL and STRING_LITERAL patterns that were in bsh-2.0.b5 (although it extends character literals to multiple characters in length).

However, this patch conflicts with the current escape syntax that is tolerant and ignores non-standard occurrences of the \ escape character in string and character literals.  For example in beanshell 2.1 "\8" seems to be tokenized as "8".  In Java the compiler would flag this escape sequence as an error.  

I am not sure of the correctness of this fix, or the intention of broadening the capture of the escape sequence.  So this might not be the way to go.  Javacc pattern matching makes my head explode.


opeongo:
After thinking about it, I really don't like the expanded escape capture.  It really isn't Java, or any other language that I know of.  It allows code that isn't Java compatible and potentially masks errors.

Also, the multi-character CHARACTER_LITERAL is pretty non-Java.  It looks a bit perl-ish.  I know that sort of alternate quote syntax is helpful for making strings that have embedded double quote characters, but is the deviation from Java worth the convenience?

Finally, I recall that a 'here data' long string literal syntax is being considered for a future Java.  I can't find the reference right now, but it would be a good idea to have the same syntax (whatever that ends up being).

Unless there is some other really good reason, I think you should consider to revert the lax escape code and the long character literal.  I will hunt around to find the mention of the long string literal and see how what is said about it.

nickl-:
Apologies bro I haven't had a chance to look at this yet. Been trying to get the next patch I've been working on committed but it just doesn't seem to want to reach an end. =/

As far as I understand the double escaping has something to do with the parser (`javacc`)

This is from java1.5 grammar

```
  < CHARACTER_LITERAL:
      "'"
      (   (~["'","\\","\n","\r"])
        | ("\\"
            ( ["n","t","b","r","f","\\","'","\""]
            | ["0"-"7"] ( ["0"-"7"] )?
            | ["0"-"3"] ["0"-"7"] ["0"-"7"]
            )
          )
      )
      "'"
  >
```

which is exctly the same as pmd grammar it seems so unofficially this would be the "official" java equivalent. 

When I did the literal revision I tried to simplify the lexer... being the bottle neck and all. Rather get the the literal out of the parser as soon as possible and then we can deal with it. 

It was at this time where I thought it would be useful to overflow char errors into string effectively turning single `'` quoted `'` multi character (larger than 1 char or in other words a failed char literal) into strings. This is rather handy especially when expressing bsh scripts within java strings and something I would like to keep. 

Still haven't looked at the patch (and now GitHub seems non responsive... running on african time) so I will conclude here for now lets hope this response commits....

My thinking... If there is a problem with it being too greedy I agree we have to fix that but I don't think going back or making the grammar more complex is the answer. We can apply more strict rules in java instead. There is not benefit in letting the parser handle the complexity. Simplify the grammar and make it easy to parse, we just need the broader tokens it doesn't need to apply the language. Just finish sharp and we'll take it from there.

Or what were you thinking?


Ahh there the internets are back... yes I thought that was what you were suggesting. At closer investigation it is quite obviously doing exactly what you explained. 

See:
```
 < CHARACTER_LITERAL: "'" (~["'","\n","\r"] | "\\'")* "'" >
|
  < STRING_LITERAL: "\"" (~["\"","\n","\r"] | "\\\"")* "\"" >
```

Bot string and char looks for `"\\'"` and will gobble it up. Not sure where this came from or what the idea was but that looks like the brokenness. From here anyway...

If I read the regex out loud:

find `'` then for 0 to many time `*` find anything that is net `~[ ]` a single quote `'` or new line `\n` or carriage return `\r` OR `|` anything that is an escaped quote `\'`... ah yes that makes sense now... until we find another single quote `'`... it should work and should allow escaped quotes right?

The escaped chars we need to parse in java anyway so no advantage letting javacc do it.

nickl-:
Strangely enough if you do it in parts it works:

```java
bsh % a='a\\b';
--> $7 = "a\b" :String
bsh % bsh % b='\\';
--> $8 = '\' :char
bsh % a.replace(b,'F');
--> $9 = "aFb" :String
```

So it would seem that the gobbler is too greedy and needs to slow down.

As mentioned before javacc escapes the escapes so what we are getting is.
```
'a\\\\b'
```

which we have to parse as 
```
{'a', '\\', '\\', 'b'}
==
{'a', '\', '\', 'b'}
==
{'a', '\'\', 'b'}
==
{'a', '\', 'b'}
==
'a\b'
```

nickl-:
So your test case works if we change remove the escaped ;quote option.

```
  < CHARACTER_LITERAL: "'" (~["'","\n","\r"])* "'" >
```

```java
bsh % 'a\\b'.replace('\\', 'F');;
--> $1 = "aFb" :String
```

But now we don't support escaped quotes anymore and the problem here is that the bracket notation only accepts single chars and it requires more than 1 char to represent an escaped quote in java string =/

nickl-:
Fixed.

I will start another discussion on what it means to work like java does but for now, this fix still accomplished the goals.

 * remove grammar complexity and improve improve parser performance
 * result should work as expected
 * result should work like java as far as possible
 * result should enhance and improve BeanShell

The grammar is still a huge improvement over the original java grammar.   
The test cases now work as expected:  

```java
bsh % 'a\\b'.replace('\\','F');
--> $0 = "aFb" :String
bsh % 'a\'b'.replace('\'','F');
--> $1 = "aFb" :String
bsh % 'a\tb'.replace('\t','F');
--> $2 = "aFb" :String
bsh % 'a\nb'.replace('\n','F');
--> $3 = "aFb" :String
bsh %
bsh % "a\\b".replace("\\","F");
--> $4 = "aFb" :String
bsh % "a\"b".replace("\"","F");
--> $5 = "aFb" :String
bsh % "a\tb".replace("\t","F");
--> $6 = "aFb" :String
bsh % "a\nb".replace("\n","F");
--> $7 = "aFb" :String
bsh %
```

Works like java, as you mentioned in a previous post the `'\8'` issue and I explained how single quotes roll over from chars to Strings on invalid chars. So instead of causing "illegal escape character" errors chars become strings.

```java
sh % '\7';
--> $8 = '' :char
bsh % '\77';
--> $9 = '?' :char
bsh % '\377';
--> $0 = 'ÿ' :char
bsh % '\8';
--> $1 = "8" :String
bsh %
bsh % "\7";
--> $2 = "" :String
bsh % "\77";
--> $3 = "?" :String
bsh % "\377";
--> $4 = "ÿ" :String
bsh %
```

So as we can see everything works like java but avoids breaking.

Enhancing BeanShell... escapes are now properly defined as 2 character literals which starts with a `\` followed by another character.

### Escaping like a boss
something rather amazing happened as a result.

```java
bsh % // string literals can't span multiple lines - like java

bsh % a = "abc

// Error: Unknown error: bsh.TokenMgrException: Lexical error at line 22, column 9.  Encountered: "\n" (10), after : "\"abc"
bsh % bsh % // but we can now escape anything =)

bsh % a = "abc\
def\
ghi\
jkl\
";
--> $5 = "abc
def
ghi
jkl
" :String
bsh % print(a);
abc
def
ghi
jkl

bsh %
```

opeongo:
> The escaped chars we need to parse in java anyway so no advantage letting javacc do it.

I am not sure what you mean by this.   Isn't there is a need to filter out invalid escaped multi-byte sequences, and provide error messages for garbage input?

> the lexer... being the bottle neck
> improve improve parser performance

I don't think there is any big performance win here.  For the common case of unescaped input both versions are pretty well the same thing, accepting the character at the first alternation.  It is only in case of an escaped sequence that that the pattern becomes more specific.  Even so, the most common of those are caught in the second alternation.

Automatically changing '\8' to "8" seems very odd to me.  In Java the former is clearly an error.  This converts to a string that is different from the original without any kind of warning.  I just don't see the merit in this change.  It makes BeanShell character behave differently than a Java primitive.

One of my concerns about adding features to BeanShell (like promotion of character string with length > 1 to a string, handling of escape sequences) is that the documentation is not also getting updated.  The nice thing about BeanShell and why I prefer it to Python or Groovy is because it is so Java-like.  Even though there isn't a lot of documentation about BeanShell it is easy to tell others that it has a Java syntax.  With each non-standard feature that is added BeanShell differentiates in an incompatible way.  

I realize the temptation to add extra features (it is easy, scratches an itch).   Can we partition these off a bit so that they are a bit separate?  Would it be possible to define a mode (similar to setStrictJava(true)) that would throw errors when a non-java feature was parsed?  How about enableAdvancedFeature()?

nickl-:
#501 was inspired, in part, by discussions in this thread.

opeongo:
I actually like the idea and the implementation of long character strings, and it is a scripting feature that I will probably use.  Your thoughts on the 'evolution' of BeanShell also help me here to understand the rational behind adding it in.

It may be that we just agree to disagree, but I still fail to see a reason why automatically converting '\8' to "8" would be something that anyone would want to do, or would expect to happen.  Isn't this acting against the principle of least surprise?  I agree that there is no need to fail like Java if there is an obvious way to recover, but in this case the correction does not seem like it is obviously better.  

If the intention in the string '\my\windows\pathname' was "\\my\\windows\\pathname", surely "mywindowspathname" is not the right answer.

And if '\m' was intended to be '\n', then surely "m" is not the right answer either.

For both of these cases, to me anyway, an exception for an invalid escape sequence would probably save me some time to figure out why my code isn't doing the right thing.  

What sort of use cases do you see for this feature?

I also understand the point of improving the performance of the lexer/parser, but looking at the pattern I don't see any potential performance gain here for most of the common cases.

nickl-:
Ok thanx for explaining at least we are on the same page. I thought you were against the added features when I took every effort that it should make sense from a java perspective.

I hear what you are saying and if anything should cause unnecessary debugging frustrations that is a valid concern. But I do feel you are making a mountain out of an ant heap.

For starters your imagined fault with string "\\my\\windows\\pathname" will in fact produce the correct result "\my\windows\pathname" and not what you expected.

in my defence ignoring false escapes in strings is not a new feature and comes from the original BeanShell string implementation which is now applied to chars too because we fail over to strings from chars.

I agree with the strategy of not failing, but perhaps loosing the `\` is not the best solution. Let me think about it, would you please do some tests too and see what the actually trade off is.

As for the grammar, sure for normal parsing it is the same but for escape parsing it is much cheaper, makes the grammar more readable and I simply love the "boss escaping". Besides throwing parser errors is not ideal either, if we have to throw an error lets make it "illegal escape character" or such, which can be accomplished from `BSHLiteral`. Would you agree?

opeongo:
I definitely agree that an "illegal escape character" exception is better than a parse exception.  This probably is worthy of an issue of it's own to discuss how to improve parser error messages.

I did misread the code and how it would treat "\\my\\windows\\pathname".  It is not what I thought.

However, I think that this entire issue should be reviewed in light of [JEP 327 "Raw String Literals"](http://openjdk.java.net/jeps/326).

Issue #444, opened 2007-12-14
Reflect.getObjectProperty() doesn't use ClassManager


sourceforge-issue-exporter:
It makes eval\(\) to be very slow

mailto:menganitoxc@gmail.com

Reported by: *anonymous

Original Ticket: "beanshell/bugs/306":https://sourceforge.net/p/beanshell/bugs/306

nickl-:
I also noticed this, there is absolutely no property cache.

Issue #81, opened 2018-07-03
New BeanShell release


nickl-:
Created on behalf of @pgiffuni

Start considering a new release (2.0?).

Which next major release version number would you prefer?

[![](https://api.gh-polls.com/poll/01CHGF0S2MPRRZV3546T759VAR/BeanShell%203)](https://api.gh-polls.com/poll/01CHGF0S2MPRRZV3546T759VAR/BeanShell%203/vote)
[![](https://api.gh-polls.com/poll/01CHGF0S2MPRRZV3546T759VAR/BeanShell%208)](https://api.gh-polls.com/poll/01CHGF0S2MPRRZV3546T759VAR/BeanShell%208/vote)

See next post for details.

nickl-:
I see two questions to this issue:

### 1. What is the next release version number?

There has been a lot of confusion in the BeanShell community since the beanshell2 fork which can be mitigated by a major release version number greater than 2 as a clear indicator of which version has preference. While contemplating this I thought of two options:

**BeanShell 3**

The most logical next increment of major release version numbers greater than beanshell 2.

**BeanShell 8**

As you may be aware the minimum JAVA requirement for the new release will be JDK 1.8.  Perhaps the major version number can also be an indicator of the minimum Java requirement. 

Hope I have polls correctly configured then perhaps we can put it up for a vote.

Which next major release version number would you prefer?

/polls "BeanShell 3" "BeanShell 8"

### 2. When can we release?

It has been several years since the last release so it would be safe to say there is no big urgency for the next release. A lot of effort has gone into this release so I would like it to be as stable as possible before opening the flood gates again. One way to ensure a stable release is through wide adoption which is not a realistic goal at the moment. Another approach, which is more attainable, would be through good test coverage and ensuring all reported issues have been addressed. 

There is a [road map](/beanshell/beanshell#development-road-map) maintained in the readme as a change log as well as checklist of things that need done before we can release. Please let us know your thoughts on this.

stain:
@nickl- are you willing to be Release Manager..? Are we keeping the groupId of `org.apache-extras.beanshell`

Let me know if you need your account added for https://repository.sonatype.org/ to be associated with that groupId -- see https://central.sonatype.org/pages/ossrh-guide.html

nickl-:
@stain Not sure if I want the official title but it is probably best if we spread the permissions around.

I added jira account nickl on sonatype, is there anything else you require?

stain:
Sorry for the delay!

[...]

PR #500, opened 2018-08-14
Tweaks

wilx:
Here are some tweaks, mostly through IntelliJ IDEA's inspection capability.


opeongo:
Some of these changes are syntactic sugar but they are making the source code dependent on Java SE >=  version 1.8 (I believe that this has already happened elsewhere in the code base, so this comment is not 100% directed at this PR).  Is this an intentional, considered choice for this project, or has it just kinda happened?  

I am not opposed to this (1.7 got EOL'ed in 2015), but it would be helpful if there were a clear statement about version requirements on the project page, a vision about version tracking and backward compatibility, and tags to identify when a dependency to a new version is introduced.  I think this would help all contributors.

wilx:
[**@opeongo**](https://github.com/opeongo) commented on [Aug 15, 2018, 12:50 AM GMT+2](https://github.com/beanshell/beanshell/pull/500#issuecomment-413040959 "2018-08-14T22:50:55Z - Replied by Github Reply Comments"):
> Some of these changes are syntactic sugar but they are making the source code dependent on Java SE &gt;=  version 1.8 (I believe that this has already happened elsewhere in the code base, so this comment is not 100% directed at this PR).  Is this an intentional, considered choice for this project, or has it just kinda happened?
> 
> I am not opposed to this (1.7 got EOL'ed in 2015), but it would be helpful if there were a clear statement about version requirements on the project page, a vision about version tracking and backward compatibility, and tags to identify when a dependency to a new version is introduced.  I think this would help all contributors.

It was mentioned in #81 in connection to version change that could be version 8 to reflect that requirement so I took it as granted.

nickl-:
> making the source code dependent on Java SE >= version 1.8

@opeongo Yes minimum 1.8 as per the build script

https://github.com/beanshell/beanshell/blob/0ad109fb38beec022adc88035d931f9a71137f8f/pom.xml#L56

There should be closed issues that references this decision... it happened a long time ago.

src/main/java/bsh/ClassGeneratorUtil.java
```diff
@@ -622,24 +622,35 @@ private void doSwitchBranch(int index, String targetClassName, String[] paramTyp
...
-            else if (type.equals("F"))
+                break;
+            case "F":
                 method = "getFloat";
-            else
+                break;
+            default:
                 method = "getObject";
+                break;
+            }
```
nickl-: Nice!  I did some benchmarks on this and switch statements really are about 75% quicker even on simple decisions like:

```java
if ( a == 1 || a == 2 )
       // do it
```

is slower than 

```java
switch (a) {
   case 1: case 2:
       // do it
}
```

src/main/java/bsh/classpath/ClassManagerImpl.java
```diff
@@ -121,8 +120,8 @@ This is lazily constructed and further (and more importantly) lazily
     private BshClassPath fullClassPath;
 
     // ClassPath Change listeners
-    private Vector listeners = new Vector();
-    private ReferenceQueue refQueue = new ReferenceQueue();
+    private Vector<WeakReference<Listener>> listeners = new Vector<>();
+    private ReferenceQueue<Listener> refQueue = new ReferenceQueue<>();
```
nickl-: I was wondering where the type parameters were. =)

The use of `Vector` vs `ArrayList` also needs to be considered here.

src/main/java/bsh/org/objectweb/asm/Frame.java
```diff
@@ -240,7 +240,7 @@
    */
```
nickl-: NO! We don't make changes to package `bsh.org.objectweb.asm.*`

src/test/resources/test-scripts/primitives_large_numbers.bsh
```diff
@@ -108,7 +108,7 @@ assertThat(t, instanceOf(Double.class));
 assertEquals(8.988465674311579E307, t, 0.0);
 t = 1.7976931348623167E616 / 2;
 assertThat(t, instanceOf(BigDecimal.class));
-assertEquals(8.9884656743115835E+615w, t);
+assertEquals(8.988465674311584E+615w, t);
```
nickl-: Huh? What's the hang up here?

Anyway... does intelij savvy `.bsh`?

src/main/java/bsh/ClassGenerator.java
```diff
@@ -73,8 +73,8 @@ public Object invokeSuperclassMethod(BshClassManager bcm, Object instance, Strin
     public static Class generateClassImpl(String name, Modifiers modifiers, Class[] interfaces, Class superClass, BSHBlock block, Type type, CallStack callstack, Interpreter interpreter) throws EvalError {
```
nickl-: I see this didn't add the wildcard type parameters for ie. `Class<?>` and `Constructor<?>` any reason why it can't?

src/main/java/bsh/Capabilities.java
```diff
@@ -136,7 +136,7 @@ public Boolean get() {
 
     @Override
     public void accept(Boolean t) {
-        this.accessibility = t.booleanValue();
+        this.accessibility = t;
```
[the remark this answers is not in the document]
nickl-: I see it passed over my example... that's sad.

src/main/java/bsh/org/objectweb/asm/Frame.java
wilx: Ooops. Agreed. Will revert.

src/test/resources/test-scripts/primitives_large_numbers.bsh
wilx: This is related to another change set where I set rounding mode for division. I am not entirely sure about that but it seems to avoid exception in case of inexact result or such.

src/main/java/bsh/ClassGenerator.java
wilx: It can but I had this particular replacement disabled to limit the diff.

opeongo:
@wilx thanks for the link to #81.  Makes sense to me now.

src/main/java/bsh/BSHAssignment.java
```diff
@@ -165,6 +165,6 @@ private Object operation( Object lhs, Object rhs, int kind )
         }
 
         throw new UtilEvalError("Non primitive value in operator: " +
-            lhs.getClass() + " " + tokenImage[kind] + " " + rhs.getClass() );
+                                lhs.getClass() + ' ' + tokenImage[kind] + ' ' + rhs.getClass() );
```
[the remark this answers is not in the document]
opeongo: I read somewhere that string concatenation gets compiled in to StringBuilder calls.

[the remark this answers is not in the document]
wilx: Yes, concatenation gets compiled into StringBuilder invocations. I think the point is the code for appending one character is significantly simpler than code appending strings.

nickl-:
@wilx Just a reminder please take note from my initial review comment:

> Your timing is slightly off though. =( I've been working on a fix for #444 which is going to touch a major part of the source base.
>
> I will give you the go ahead once I'm done.

Whatever you are doing now will have to be done over so best give me chance to get this patch in before you go on. It should not take very much longer now... about the length of a piece of string =)

wilx:
[**@nickl-**](https://github.com/nickl-) commented on [16 Aug 2018, 08:47 CEST](https://github.com/beanshell/beanshell/pull/500#issuecomment-413443364 "2018-08-16T06:47:30Z - Replied by Github Reply Comments"):
> [@wilx](https://github.com/wilx) Just a reminder please take note from my initial review comment:
> 
> > Your timing is slightly off though. =( I've been working on a fix for [#444](https://github.com/beanshell/beanshell/issues/444) which is going to touch a major part of the source base.
> > 
> > I will give you the go ahead once I'm done.
> 
> Whatever you are doing now will have to be done over so best give me chance to get this patch in before you go on. It should not take very much longer now... about the length of a piece of string =)

OK.

PR #506, opened 2018-08-21
Raw string literals like JEP 327

opeongo:
This patch implements raw string literals like in [JEP 327 "Raw String Literals"](http://openjdk.java.net/jeps/326).  This JEP is scheduled to land in Java 12.


nickl-:
@opeongo awesome stuff ty for taking the initiative.

Will need to look into this some more but kinda feel we are overdoing it, do we really have to count the ticks? Isn't there a way to simply catch them with a regex?

nickl-:
> The following regexp should work: `` (`++)(.+?)\1 `` with the resulting raw String being in group 2. Note that this doesn't process line endings and it's questionable if this is faster than the "manual" way.

We need to translate that to a javacc but yes it will definitely be faster than counting opens and closes manually. If we could accomplish this for parameter types this must be much easier.

Issue #72, opened 2018-06-18
Documentation Writing unit tests


nickl-:
I configured coverage reports with the CI build, you may have noticed the new badge on the project front page: [![codecov](https://codecov.io/gh/beanshell/beanshell/branch/merge-fork-beanshell2/graph/badge.svg)](https://codecov.io/gh/beanshell/beanshell)

I would like this to reach a more acceptable percentile before we roll the release. Please assist if you have some time available so we can wrap this up.

There are two ways to contribute unit tests.

**The new old bsh test scripts**: This has become my favourite platform, no need to eval scripts from java strings. Just script bsh like it was meant to. Many new features like junit asserts and hamcrest matchers, improved error reporting and running single script tests has brought it on par with a modern testing framework.

To get started you can copy. paste and rename the [template script](https://github.com/beanshell/beanshell/blob/12678efd04431fdd5d74d6dadca927af23031e2d/src/test/resources/TestTemplate.bsh) to `src/test/resources/test-scripts/` and run your script with the mvn command line:

```shell
$ mvn test -Dtest=BshScriptTestCase -Dscript=your_test_script.bsh
```

The template is documented and you can look at some the newer test scripts as examples.

Or if you prefer...
**The good old faithful junit tests**: You can copy, paste and rename one of the existing tests, into `src/test/java/bsh/`. Some of the newer tests may make better templates, try [TryStatementTest](https://github.com/beanshell/beanshell/blob/12678efd04431fdd5d74d6dadca927af23031e2d/src/test/java/bsh/TryStatementTest.java), [FinalModifierTest](https://github.com/beanshell/beanshell/blob/12678efd04431fdd5d74d6dadca927af23031e2d/src/test/java/bsh/FinalModifierTest.java) or [TypeParametersTest](https://github.com/beanshell/beanshell/blob/12678efd04431fdd5d74d6dadca927af23031e2d/src/test/java/bsh/TypeParametersTest.java), these all use  the`TestUtil` convenience class. 

Surefire (maven junit test runner plugin) includes the following names by default which you may consider when naming your tests:

`**/Test*.java` - includes all of its subdirectories and all java filenames that start with "Test".
`**/*Test.java` - includes all of its subdirectories and all java filenames that end with "Test".
`**/*TestCase.java` - includes all of its subdirectories and all java filenames that end with "TestCase".

You can single out your test with the following command:

```shell
$ mvn test -Dtest=TestClassName
```

stefanofornari:
[...]

PR #509, opened 2018-08-21
Test for array post fix operator

opeongo:
This patch add a test for postfix array operators.  There does not appear to be any problem, just adding to the test coverage.


nickl-:
Awesome! as per #72 tyvm

Issue #511, opened 2018-08-24
Error referencing local variable


opeongo:
An error occurs when referencing a local variable within a class.  The following code will throw a 'no such field' exception during the assignment to one.ref:

```java
public class TestAccess {

   public TestAccess ref = null;

   public static void main(String[] args) {
      TestAccess one = new TestAccess();
      one.ref = one;
   }
}
```

I tried to track down where the problem was happening, and noticed an exception occuring in ClassGenerator.java at line 187, where tvd.evalType happens.  The error is ignored and the variable does not get defined.
```java
                BSHTypedVariableDeclaration tvd = (BSHTypedVariableDeclaration) node;
                Modifiers modifiers = tvd.modifiers;
                BSHVariableDeclarator[] vardec = tvd.getDeclarators();
                for (BSHVariableDeclarator aVardec : vardec) {
                    String name = aVardec.name;
                    try {
                        Class<?> type = tvd.evalType(callstack, interpreter);
                        Variable var = new Variable(name, type, null/*value*/, modifiers);
                        vars.add(var);
                    } catch (UtilEvalError | EvalError e) {
                        // value error shouldn't happen
                    }
                }
```

[TestAccess.java.txt](https://github.com/beanshell/beanshell/files/2318983/TestAccess.java.txt)

nickl-:
There still needs to be more work done separating properties from fields, yes.

nickl-:
This appears to be resolved:

```java
bsh %
public class TestAccess {

   public TestAccess ref = null;

   public static void main(String[] args) {
      TestAccess one = new TestAccess();
      one.ref = one;
   }
}
--> $2 = class TestAccess :Class
bsh % TestAccess.main(new String[0])

--> void
```

**Closed: Fixed**

PR #512, opened 2018-08-27
Java classes that reference themselves

opeongo:
This PR is related to #511.

There is a problem when creating Java classes that have a local variable that references the enclosing class. This is a common idiom for data structures and used to work under bsh-2.0.b5.  

The problem seems to occur in the class generator.  The class is not defined yet when trying to look up the self-referential declared variable type.  In the current code base the exception is ignored and the declared variable is simply ignored.  Later on an error will occur when there is an attempt to reference this 'undefined field'.  

This patch will use the class type descriptor when the class is not available, and this will defer resolution of the class until later.

This patch solves the problem identified in the test script for #511.  However, as a side effect it has broken `bsh.EnumTest.enum_values_excludes_enum_static_field(EnumTest.java:137)`.  I am not sure exactly how to fix this other one, as I suspect it has been 'uncovered' by this fix for the self reference issue.  I offer this PR up as a starting point to identify the problem.


opeongo:
I had a look in to why the `bsh.EnumTest.enum_values_excludes_enum_static_field` test failed with my patch applied.  Here is what the failing test was doing:

```java
    @Test
    public void enum_values_excludes_enum_static_private_field() throws Exception {
        try (final Interpreter bsh = new Interpreter()) {
            Object[] obj = (Object[]) bsh.eval(script(
                "enum E4 {",
                    "VAL1, VAL2, VAL3, VAL4;",
                    "private static final E4 enm = E4.VAL1;",
                "}",
                "E4.values();"
            ));
            assertThat("array with length length is 4", obj, arrayWithSize(4));
            assertThat("array containing VAL1, VAL2, VAL3, VAL4", obj, arrayContaining(
                    bsh.eval("E4.VAL1"), bsh.eval("E4.VAL2"),
                    bsh.eval("E4.VAL3"), bsh.eval("E4.VAL4")));
        }
    }
```
The test failed because E4.values() returned 5 elements:  this included the 4 enum values plus the static field enm.  It looks like prior to my patch that the declaration of enm was silently ignored, so the enum class only had four fields.  Prior to the patch a test for `E4.VAL1 == E4.enm` would have failed with enm being an undefined field.

The reason that 5 elements were returned was because the synthetic method to get the enum values simply checks for static fields that are of the same class as the enum.  This method captures the statically defined enm field as well as all the value fields.
```java
    static <T> T[] getEnumConstants(Class<T> enm) {
        return Stream.of(enm.getFields())
                .filter(f -> f.getType() == enm)
                .map(f -> {
            try {
                return f.get(null);
            } catch (Exception e) {
                return null;
            }
        })
        .filter(Objects::nonNull)
        .toArray(len -> (T[]) Array.newInstance(enm, len));
    }
```
Fortunately there is a field reflection method that identifies enumeration constants, and it is unsurprisingly named `isEnumConstant`.  When I add this to the map test it seems to solve the problem in a very simple way.  

Patch for this plus updated test pushed

nickl-:
The fields vs properties issue... will need to investigate.

PR #515, opened 2018-09-02
Banner in main

stefanofornari:
This patch is a suggestion to move the printing of the banner to main(). Currently it is in the run() method which is actually used in many other cases where the banner is not supposed to be displayed.

Feel free to trash this PR if you do not like it.

PS: sorry for the multiple commits... my workflow is still not great :(


nickl-:
you can squash commits with `git rebase --interactive` 

Will have a look at this... tyvm

stefanofornari:
I will create a new one from scratch ...

stefanofornari:
replaced by #518

Issue #522, opened 2018-09-27
division by zero error when multiplying 0 value


HaasJona:
Multiplying zero with a number can lead to a division by zero error:

```
    @Test
    public void testZero() throws Exception {
        final PreparsedScript f = new PreparsedScript(
                "return 0 * 2;",
                _classLoader
        );
        assertEquals(0, f.invoke(Collections.emptyMap()));
    }
```

Expected result: Test passes.

Actual result:
```
bsh.TargetError: Failed operation: 0 "*" 2: Arithemetic Exception in binary op : at Line: 1 : in file: inline evaluation of: ``__execute() { return 0 * 2; } return this;'' : ; 

Called from method: __executeCaused by: java.lang.ArithmeticException: / by zero

	at bsh.UtilTargetError.toEvalError(UtilTargetError.java:62)
	at bsh.BSHBinaryExpression.eval(BSHBinaryExpression.java:106)
	at bsh.BSHAssignment.eval(BSHAssignment.java:42)
	at bsh.BSHReturnStatement.eval(BSHReturnStatement.java:42)
	at bsh.BSHBlock.evalBlock(BSHBlock.java:145)
	at bsh.BSHBlock.eval(BSHBlock.java:89)
	at bsh.BshMethod.invokeImpl(BshMethod.java:425)
	at bsh.BshMethod.invoke(BshMethod.java:297)
	at bsh.BshMethod.invoke(BshMethod.java:201)
	at bsh.PreparsedScript.invoke(PreparsedScript.java:77)
	at bsh.PreparsedScriptTest.testZero(PreparsedScriptTest.java:75)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #523, opened 2018-10-03
Fix issue 522 

jmetertea:
Fixed https://github.com/beanshell/beanshell/issues/522
Prevent two cases of division by zero error


nickl-:
Perfect tyvm

Issue #538, opened 2019-02-27
Add an optional variable missing hook like invoke() [proposed feature]


drosenbauer:
**Edit:** documentation from discussion at [issuecomment-474192792](/beanshell/beanshell/issues/538#issuecomment-474192792)

When writing Beanshell code, the following standard boilerplate is quite common:

```java
if (a != void && a != null) { 
    // do something with a
}
```

For methods, any namespace can define an `invoke(name, args)` method that will be called for any missing method declaration. I propose a similar catch-all hook for variables.

My proposed syntax is:

```java
Object missingVariable(String name) {
    // return something 
}
```

This method, if it exists, would be invoked by the Interpreter in `getVariable` if a matching variable does not exist or is equal to `Primitive.VOID`.

nickl-:
I must be missing something but I don't see the point.

For what purpose would you need a default value for undefined variables?

Even this is weird:

```java
if (a != void && a != null) { 
}
```

What do you expect `a` to be? 

If `a` is typed then it can't be `void` because it will by definition be defined.
So `a` must be untyped but where did it come from?
Surely the whole purpose of the statement is to ensure `a` was assigned but then it cannot possibly be `void`.
You cannot assign `void` or pass `void` as an argument.

> // Error: Evaluation Error: Undefined argument
> // Error: Evaluation Error: Void assignment.

I am really trying to wrap my head around this but I fail to see the point. Please excuse me.

Could you perhaps give me a use case?

drosenbauer:
Scripting with bindings. My main use of Beanshell is as a scripting language within another product. It's not always clear what will be passed to my scripts as bindings, so anytime I need to use an uncertain variable, I need to check both void and null. I can't even just check null, because the null check fails on voids.

It's the same reason Groovy has `propertyMissing` and Nashorn gave Javascript `__noSuchProperty__`.

nickl-:
Ahh makes more sense with a use case, thanks!

I took some time to produce some more documentation just to clarify why the following information may seem generalised they should proof useful for understanding exactly what `void` means and what we can use it for.

## Loose typed

BeanShell is a loose typed scripting language, each value has a type but variables may be assigned any value and methods may return any value.

Here is a variable example being assigned different typed values:

```java
bsh % a = "I'm a string";
--> $7 = "I'm a string" :String
bsh % a = 123;
--> $8 = 123I :int
bsh % a = 456L;
--> $9 = 456L :long
```

This example shows a method returning different typed values:
```java
[...]

PR #540, opened 2019-02-27
Add two mechanisms for dealing with voids

drosenbauer:
Added two mechanisms for dealing with `void`s in Beanshell: a propertyMissing(name) method that will be invoked when any variable cannot be resolved and an optional Capability to transform all void values to nulls during evaluation.

For issue #538


nickl-:
Can you please elaborate.

Issue #537, opened 2019-02-25
Clarify the version situation


drosenbauer:
You've got [2.1.8](https://code.google.com/archive/p/beanshell2/downloads) over here on the old Beanshell2 Google Code site and [2.1.9](https://github.com/pejobo/beanshell2/tree/master/dist) over here on this fork. Since Beanshell was inactive for so long, there are a number of commercial products ((example)[(https://support.ca.com/cadocs/0/CA%20DevTest%20Solutions%208%200-FRA/Bookshelf_Files/HTML/Third_Party/index.htm?toc.htm?2337400.html)) referring to their Beanshell version as 2.1.8.

We've got 2.0b6 over here in this branch. 

There's also a 1.3.0, which appears to the uninitiated eye be the most recent *stable* version that isn't a fork - and that's from 2007.

And of course, 2.1 is bigger than 2.0. If I knew nothing about the fork-merge situation, I would absolutely choose 2.1.9 over 2.0b6 or the built-from-source 2.1-SNAPSHOT. That obviously puts a damper on this project but also may put users at risk if a security issue is fixed in a future 2.0b7 but not a future 2.1.10.

Please consider tagging this development branch as 2.2-SNAPSHOT, which would clarify everything.

nickl-:
@drosenbauer my sincere sympathy for your frustration I understand all too well.

It would appear cadocs is using pejobo/beanshell2 since they are still on GPL whereas we are under the Apache license. If you're not confused enough yet the last pejobo release was [2.1.7](/pejobo/beanshell2/releases) =)

But it really doesn't matter, we are working from the development branch `merge-fork-beanshell2`. This is the only code base we currently support. All other releases are end of line and unsupported. Rest assure, whatever security issues and bugs yet discovered will be fixed against HEAD. There is no turning back anymore...

This is also the recommended version as we cannot in good conscience suggest anyone use any of the public releases. Unless you've compiled the source yourself you are NOT using the preferred version.

I understand your concerns and they are all valid. My big concern is the lack of exposure as no one is currently using / testing what we have. Also we desperately need more bodies, can barely cope with the mundane. There has been many strides made over the last year but we are not quite there yet. You talk about security concerns, we have a backlog of 240 issues imported from the old sourceforge project. I've barely skimmed the surface. Talk about concerns...

[...]

PR #542, opened 2019-02-27
Update version to 2.2.0-SNAPSHOT per #537

drosenbauer:
I really do think this is better, as it makes it clear that this version is newer vs. beanshell2 @ 2.1.9.


nickl-:
We were actually contemplating whether to go BeanShell3 or if we should rather go BeanShell8 where was that post...

nickl-:
Issue #81 so I think we should rather go that route.

nickl-:
Thanks but we're not ready for version bump yet

Closed: Not Ready

Issue #132, opened 2000-08-29
Method selection with null arguments is not right....


sourceforge-issue-exporter:
Method selection doesn't catch when selections with null arguments are ambiguous... instead it pushes selection to the most derived match.

Reported by: patn

Original Ticket: "beanshell/bugs/11":https://sourceforge.net/p/beanshell/bugs/11

nickl-:
**Closed: Resolved as per unit tests #552**

opeongo:
In Java 16 String.valueOf(null) will call new String(char[]).

nickl-:
> Unfortunately BeanShell only uses the type of the object being passed and does not look at the declared type of the variable.

Correct since there is no guarantee that it was declared of type.

This seems difficult, not sure we even have the variable declaration type at the point of method resolution.

This issue was regarding NullPointerException being thrown, and if I understand your analysis correctly we now return null without exception, is this correct? If so, does it really matter which method handled the null? ...as long as the non null values are handled correctly.

If we are not causing NullPointerException anymore then this issue can be closed, yes? Oh I see it is closed...hmmm.

opeongo:
The problem is that at the moment a null argument cannot carry any type, so when beanshell does the method selection it is prone to having an ambiguous method selection problem that it resolves differently than Java.  

The `String.valueOf()` example is an easy on to look at because there are only two method options:
```java
String.valueOf(Object obj)
String.valueOf(char[] value)
```

If there is just a plain unadorned null with no type then according to the method selection rules `char[]` is more specific than `Object`, so the `String.valueOf(char[] value)` method should be selected.  This will throw an NPE when passed a null.  This is what Java does.

However, the deck is stacked [by these lines](https://github.com/beanshell/beanshell/blob/cb552a4d08c279680d6463b61e2f034fe12a590b/src/main/java/bsh/Types.java#L310-L313) in BeanShell that say if the argument is null then give it the type of String.class.  This is a weird trick to always make nulls match to String arguments.  As you can expect it works occasionally but fails just as often.  Because of this `String.valueOf(null)` always returns `"null"`.

If this was Java (statically typed) then the compiler most often than not would know the type and be able to make the correct method selection:
```java
Object a = null;
assert(String.valueOf(a).equals("null"));
assert(String.valueOf((Object)null).equals("null"));
```

[...]

PR #552, opened 2019-05-28
Add test for method selection with null parameter - identified in Iss…

opeongo:
…ue #132.

This was probably fixed when Issue #488 was resolved.  Issue #132 no longer seems to be a problem.


nickl-:
@opeongo nice one tx!

Issue #234, opened 2002-12-16
Control-d and Delete key problems


sourceforge-issue-exporter:
Using the latest BeanShell desktop 1.0,
On my Bsh Workspace, pressing control-d
does not cause it to exit at all.  Nothing
happens.

On the same Bsh Workspace, the delete key
does not delete if the caret is the position 0 of
the buffer.  For example,
You enter:
Pwd\(\);

Then you realize that it should be:
pwd\(\);

You move the cursor to the P position with the
purpose to delete P and then enter p.  This
does not work.

Binh

PS: JDK is 1.4.0-rc and my OS is win 2000.

Reported by: *anonymous

Original Ticket: "beanshell/bugs/96":https://sourceforge.net/p/beanshell/bugs/96

nickl-:
@opeongo Can you make out what exactly the **first part of this bag** is supposed to be?

opeongo:
The first part of the bug report is that pressing the Ctrl-d key does not close the BeanShell Workspace window.  

I don't think that this is necessarily a bug.  Different OS use different key presses to indicate EOF in their terminal handlers.  Ctrl-Z on some systems, Ctrl-D on others.  Why should the GUI need to react to these legacy key presses?  If you want to close the GUI then use the GUI metaphors and click on the window close icon, or right-click and select Exit.

My personal opinion is not a bug.

nickl-:
> I don't think that this is necessarily a bug. Different OS use different key presses to indicate EOF in their terminal handlers. Ctrl-Z on some systems, Ctrl-D on others. Why should the GUI need to react to these legacy key presses? If you want to close the GUI then use the GUI metaphors and click on the window close icon, or right-click and select Exit.

Agreed, key combinations are propriety it's like trying to make `Alt+F4` consistent across operating systems.

Will review PR for fixing the backspace issue and close this on merge.

nickl-:
Fixed at #553

PR #553, opened 2019-05-29
Fix for Issue #234, allow delete at start of line.

opeongo:
Please consider accepting this patch that addresses the second part of issue #234


nickl-:
See comment #234

leginee:
In order to get accepted, this PR needs to add unit test for the test coverage?

opeongo:
This fix is for a visual aspect of  an interactive component.  I am not sure how you would set up a test.  Certainly there is no capability for this in the existing test harness.

nickl-:
This one slipped through the cracks and then life happened.

PR #562, opened 2019-07-14
Performance: O(1) Listener add/delete

morris821028:
This is a simple test case for performance turning.

The origin type `Vector` will use O(n) time on `removeElement`.

```java
package bsh;

import java.util.Scanner;

public class PerformanceStateTest {
	public static String getName(String s) {
		return s + "A";
	}
	
	public static void main(String[] args) {
		final Interpreter interpreter = new Interpreter();

		Runtime runtime = Runtime.getRuntime();

		long usedMemoryBefore = 0;
		long usedMemoryAfter = 0;
		long start;
		long end;
		
		Scanner cin = new Scanner(System.in);
//		
		cin.next();
		
		try {
			interpreter.eval("import bsh.PerformanceStateTest;");
			interpreter.eval("a = \"1\" + \"2\";");
			interpreter.eval("String getName(String s) { return \"\" + PerformanceStateTest.getName(s); }");
			interpreter.eval("String getName2(String s) { if (s.charAt(0) == '1') return s + \"B\"; }");
			interpreter.eval("b = \"3\" + \"4\";");
		} catch (EvalError e) {
			e.printStackTrace();
		}
		
		start = System.currentTimeMillis();
		try {
			String v = null;
			for (int i = 0; i < 500000; i++) {
				v = (String) interpreter.eval("getName(\"123\")");
			}
			System.out.println(v);
		} catch (EvalError e) {
			e.printStackTrace();
		}
		end = System.currentTimeMillis();
		final long goldTime = end - start;
		System.out.printf("time %d", goldTime);
	}
}
```

### DIFF
--- a/src/main/java/bsh/classpath/ClassManagerImpl.java
+++ b/src/main/java/bsh/classpath/ClassManagerImpl.java
@@ -33,11 +33,13 @@ import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.util.Collection;
-import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
-import java.util.Vector;
+import java.util.Set;
 
 import bsh.BshClassManager;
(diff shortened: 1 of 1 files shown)


nickl-:
@morris821028 Nice 1 Thanx

PR #571, opened 2019-10-13
Fix travis build

matthinc:
I recognized during my last PR, that the travis-ci build fails at `jdk_switcher`.
Luckily this can be fixed by adding just one line to .travis.yml


matthinc:
EDIT - added this fix to my last PR to fix build. This one can be closed or merged separately.

Issue #485, opened 2005-12-28
Interpreter put


sourceforge-issue-exporter:
Hi i use method put of Interpreter class for pass java
class instance to bean shell. How i limit to use only
public method of java instance class?
Now i can invoke olso private and protect method.

thanks

Reported by: webdel

Original Ticket: "beanshell/support-requests/5":https://sourceforge.net/p/beanshell/support-requests/5

opeongo:
I am not able to reproduce this problem.  I have added a test case to see that this works as expected.

nickl-:
Tx @opeongo 

**Closed: resolved**

PR #573, opened 2019-11-28
Make sure that private access tests are run - #485

opeongo:
This patch makes sure that the method and field access tests are run.  This relates to issue #485


opeongo:
Mistakenly committed my changes to master.  Will make pulls of separate issues

PR #574, opened 2019-11-29
Make sure that private access tests are run

opeongo:
This patch will change the accessibility test script to make sure that a test for no access to public and package protected fields and methods is allowed.  The previous test script would not perform these tests
if the haveAccessibility capability was not present, but this capability is not required for the test.


nickl-:
Assuming this replaces #573 and relates to issue #485

Source looks good RTBC if travis build passes.

PR #575, opened 2019-11-29
Fix NPE when doing "import *;"

opeongo:
Move the accessor for ambigName to after the check for superImport. If doing a superImport the ambigName access will cause an NPE, yet the ambigName is not required for this case.


nickl-:
@opeongo Can we have only the lines changed in this commit, please.

opeongo:
Do you mean without the whitespace changes?  If this is what you are looking for then when you [review the 'files changed tab' up above](https://github.com/beanshell/beanshell/pull/575/commits/37b3a05ba382292f63333156838cf5d4bb089f08) there is a settings button that will let you ignore whitespace.

Otherwise I'm not sure what you are asking about.

Issue #556, opened 2019-06-08
EvalError vs dynamic language exception causes


nickl-:
The purpose of `EvalError` is designed to abort script when syntax or language errors occur which are considered unrecoverable. These are meant to include the JAVA equivalent of conditions which normally result in compile time errors.

There were arguments raised by @drosenbauer and @opeongo that while BeanShell aims to be a dynamic scripting language it enforces unnecessarily strict conditions by conforming to all the same language parameters as JAVA does. There may be some of the static typed exceptions which should not apply to a dynamic typed language in general.

BeanShell affirmed its policy on the JAVA likeness #501 which clarified that while BeanShell aims to "work like java" it must not be mistaken to also mean that BeanShell will "break like java does". Which implies that should BeanShell not fail like JAVA for the conditions of a certain compile time error it would not be in violation of the policy including where `StrictJava` compliance is concerned.

For the purpose of finding an appropriate BeanShell centric solution the scope of the problem needs to be clearly defined by producing a comprehensive list of all the causes that previously were considered `EvalError`, identifying clear distinctions and associate appropriate types which may lead to the allocation of the appropriately typed dynamic language exceptions.

The initial analysis of the previous errors can be collected at the [BeanShell EvalError list](https://docs.google.com/spreadsheets/d/1YWOUM6oYDUqxlSG5m7A7Sj6B00pan4B4GM6AqlSO3NA/edit?usp=sharing) google docs sheet. Anyone who wants to assist can email me for write access.

nickl-:
Updated the [working spreadsheet](https://docs.google.com/spreadsheets/d/1YWOUM6oYDUqxlSG5m7A7Sj6B00pan4B4GM6AqlSO3NA/edit?usp=sharing) to include all `throw new EvalError` instances in source code.

`UtilEvalError` also results in `EvalError` and needs to be mapped as well. Some of the non AST eval errors thrown should probably be `UtilEvalError` and needs to be investigated.

`ReflectError` as produced by `Reflect.java` also appears to result in `EvalError` which needs to be investigated as well.

opeongo:
This is not so bad, your spreadsheet lists that there are only 54 of them.  This is not unmanageable.

Is the issue here to go through this list and for each case determine if the error is truly non-recoverable, or if it is not so bad and that if caught then the intepreter could continue?  Do I understand this correctly?

If so, then we will need a new type of Exception that is less serious than EvalError that is catchable within scripts.  If we could start from scratch I would suggest that we use
[...]

PR #577, opened 2019-11-29
Handle the situation of a NoClassDefFoundError during class loading.

opeongo:
Let the NoClassDefFoundError class loader error fall through so that BeanShell can continue with it's loading efforts.  If the class cannot be loaded than an EvalError will eventually be thrown.

I don't know if this patch is doing the right thing (there is no right answer here), but I think for Window anyway it is better than the current situation.  At least with this patch the error will get wrapped with an EvalError.

Related to #556, I think that this shows a case where there is a use for a recoverable error type.  I can see in this situation that there could be cases where is an attempt to load a class in a script.  The class is not found, and if the error could be caught then the script could go on with plan B.

### DIFF
--- a/src/main/java/bsh/classpath/ClassManagerImpl.java
+++ b/src/main/java/bsh/classpath/ClassManagerImpl.java
@@ -247,13 +247,13 @@ public class ClassManagerImpl extends BshClassManager
                 // fall through
 */
             } catch ( NoClassDefFoundError nce ) {
-               
+
                /*
                 * This error can happen when the internal class name has a different
                 * spelling or capitalization than the external name of the resource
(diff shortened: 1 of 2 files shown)


opeongo:
I see that my pull request failed with check-style errors.  

How can I set up maven on my end to run these same checks prior to me making the pull request?  It seems like this would save some round-trips.

nickl-:
> I see that my pull request failed with check-style errors.
> 
> How can I set up maven on my end to run these same checks prior to me making the pull request? It seems like this would save some round-trips.


It should run by you as well, will have to check when check style runs but try `mvn install`.

PR #580, opened 2020-01-09
NPE reporting fix

### LINKED ISSUE #572: Beanshell via BSF unhelpful Null Pointer Exception in line 0 issue
Hi

I found when using BeanShell via BSF we get unhelpful error messages, such as 

> Null Pointer Exception in line 0

In the stack trace I don't see a lot of information to help me track down the problem.  I found the problem was with the  bsh.util.BeanShellBSFEngine class, attached is my fix for this:

[BeanShellBSFEngine.zip](https://github.com/beanshell/beanshell/files/3871592/BeanShellBSFEngine.zip)

I'm currently using this fix and it reports the exact line number and code details of the issue, hope this helps.

christiancairney:
Beanshell emitting NPE's via the BSF integeration just throws an error reporting a "NullPointerException" in line 0.

This fix adds meta data to the NPE giving the line number and the offending line.

Fixes #572

### DIFF
--- a/bsh-bsf-engine/src/main/java/bsh/util/BeanShellBSFEngine.java
+++ b/bsh-bsf-engine/src/main/java/bsh/util/BeanShellBSFEngine.java
@@ -52,7 +52,9 @@ public class BeanShellBSFEngine extends BSFEngineImpl
         try {
             interpreter.set( "bsf", mgr );
         } catch ( EvalError e ) {
-            throw new BSFException(BSFException.REASON_OTHER_ERROR, "bsh internal error: "+e, e);
+        	BSFException bsf = new BSFException(BSFException.REASON_OTHER_ERROR, "bsh internal error: "+e, e);
+        	bsf.initCause(e);
+            throw bsf;
         }
 
         for(int i=0; i<declaredBeans.size(); i++)
@@ -83,7 +85,9 @@ public class BeanShellBSFEngine extends BSFEngineImpl
             try {
                 object = interpreter.get("global");
             } catch ( EvalError e ) {
-                throw new BSFException(BSFException.REASON_OTHER_ERROR, "bsh internal error: "+e, e);
+                BSFException bsf = new BSFException(BSFException.REASON_OTHER_ERROR, "bsh internal error: "+e, e);
+                bsf.initCause(e);
+                throw bsf;
             }
 
         if ( object instanceof bsh.This )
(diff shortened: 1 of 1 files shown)


bsh-bsf-engine/src/main/java/bsh/util/BeanShellBSFEngine.java
```diff
@@ -52,7 +52,9 @@ public void initialize ( BSFManager mgr, String lang, Vector declaredBeans)
         try {
             interpreter.set( "bsf", mgr );
         } catch ( EvalError e ) {
-            throw new BSFException(BSFException.REASON_OTHER_ERROR, "bsh internal error: "+e, e);
+        	BSFException bsf = new BSFException(BSFException.REASON_OTHER_ERROR, "bsh internal error: "+e, e);
+        	bsf.initCause(e);
```
nickl-: This is redundant as we are already passing `e` to the constructor.

christiancairney: Happy to remove this, however the BSFException(String , Throwable) does not chain exceptions.  AFAIK the BSFException class constructor:

 public BSFException (int reason, String msg, Throwable t) {
	this (reason, msg);
	**targetThrowable = t;**
  }

targetThrowable isn't chained, and when the exception is thrown the only way to get they caused by is to use the method:

**public Throwable getTargetException ()**

This is the reason why I've initCaused(throwable) so the exception becomes chained.

christiancairney: Happy to make this change, although the reason why this done is to ensure the BSFException supported Exception chaining.  

AFAIK The BSFException class does not populate the cause of the exception use the constructor:

public BSFException (int reason, String msg, **Throwable t**) {
    this (reason, msg);
    **targetThrowable = t;**
  }

Instead it tracks the throwable via a class variable which is exposed using the getTargetException() method.

Initialising the cause ensures they getCause() returns the expected, removing the initCause() stopped exception chaining working for me.

christiancairney: @nickl- I'll hold off making the change until you've commented on the exception chaining issue and if the desire is to keep exception chaining disabled.  BR.

nickl-: I see well spotted!! `initCause` will be the desired route then.

PR #587, opened 2020-05-04
scripts/bsh: Grep correct bsh.Interpreter class signature

nomadium:
It looks like in the past bsh.Interpreter extended another class but that's not correct anymore. To grep just the class name in bsh shell scripts should be a more future proof approach.


jimjag:
Committed to master and v2 branch

Issue #590, opened 2020-09-02
An arithmetic bug about addition (+) and multiplication (*)?


Minjian:
I update my project from `bsh-2.0b4.jar` to `bsh-3.0.jar`, then some test scripts are failing.

I tested the following codes:
```
  int value1 = 1000 + 60 * 0;
  debugPrintln("debug value1 = " + value1);
  int value2 = 60 * 0 + 1000;
  debugPrintln("debug value2 = " + value2);
  int value3 = 1000 + (60 * 0);
  debugPrintln("debug value3 = " + value3);
```
Then I got this interesting output:
```
  debug value1 = 0
  debug value2 = 1000
  debug value3 = 1000
```

For `value1`, looks like it evaluates `1000 + 60` first and doesn't follow the correct arithmetic order for calculation. I fixed my test failures by using parenthesis like what I did for `value3`. Does anyone have similar issues in their project?

opeongo:
Yes it looks like operator precedence was undone with commit 8e712bad002f33e5238dd00a0cb7c2ba2760322e.  This commit needs to be reverted.

opeongo:
Pull request #606 that fixes this problem

nickl-:
Reverting commits is not fixing issues but discarding the fixes already made. Instead of looking for the commit that changed it, rather find the problem and solve it.

opeongo:
Looking at the javacc grammar it is obvious that the standard math operator precedence rules were removed by the original  8e712bad002f33e5238dd00a0cb7c2ba2760322e commit.  In additional I added a set of test case that failed with the original commit, and succeeded after the revert.

Also, why remove the test cases that I added?  They show valid expressions and expected results.  They should be left in, as with all tests, to validate behaviour and detect regressions.

elbosso:
I second it - please revert this optimization from hell as it screws up virtually everything and makes the library virtually unusable! I am astonished that it takes this long a discussion - the commit breaks a central language feature and should have been reverted days after the issue was raised!

elbosso:
I just saw that the pull request got merged half a year ago - how about closing this issue then?

opeongo:
> I just saw that the pull request got merged half a year ago - how about closing this issue then?

Hi @elbosso.  My opinion is that the commit needs to be reverted again before the issue can be closed.  However @nickl- is the one who made the original commit and who undid my reversion, and he seems to have a strong opinion on the matter.  I will leave it up to him to weigh in on how this problem should be solved.

[...]

Issue #568, opened 2019-08-13
Unable to Parse Ternary Operator Expression in Function Argument


morris821028:
## Problem ##

```
_func() {
    a = new Integer(1);
    return String.format("%s" + "a" + "%s", a == null ? "3" : "4", a == 1 || a == null ? "5" : "6");
}

System.err.printf(_func());
```

## Env ##

* BeanShell 3.0.0-SNAPSHOT.841

## Expect Result ##

```
4a5
```

Reference https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/javacc/java.jj#L1429

```java
package bsh;

import org.junit.Assert;
import org.junit.Test;


public class TernaryOperatorTest {
	static String _func() {
		Integer a = 1;
		return String.format("%s" + "a" + "%s", a == null ? "3" : "4",
				a == 1 || a == null ? "5" : "6");
	}

	static final String funcScript = "String _func() {\n" + "    Integer a = 1;\n"
			+ "    return String.format(\"%s\" + \"a\" + \"%s\", a == null ? \"3\" : \"4\", a == 1 || a == null ? \"5\" : \"6\");\n"
			+ "}";

	public static void main(String[] args) {
		System.err.println(funcScript);
		System.err.printf("%s\n", _func());
	}

	@Test
	public void parseInArgument() throws Exception {
		final Interpreter interpreter = new Interpreter();

		interpreter.eval(funcScript);
		String v = (String) interpreter.eval("_func()");
		Assert.assertEquals("4a5", v);
	}
}
```

nickl-:
@morris821028  Well spotted!

This test can be simplified:

```java
print(true || false ? "True" : "False");
```

or even simpler still:

```java
res = true || false ? "True" : "False";
```

This is a parser error and needs to be fixed in grammar.

As a workaround placing the condition in brackets works.

```java
print((true || false) ? "True" : "False");
``` 

But since this works in kava it should also work the same in BeanShell.

nickl-:
This may be a simpler root cause for the issue raised at #390

opeongo:
This broke due to the removal of operator precedence in 8e712ba, and is fixed in PR #671

nickl-:
Closed works as expected

PR #606, opened 2021-05-24
Revert "Grammar optimisation reducing 17 productions down to 4."

opeongo:
Here is a fix for issue #590 as well as a more complete fix for issue #568.  Both of these were caused by the operator precedence rules bring dropped in an earlier commit.

This reverts commit 8e712bad002f33e5238dd00a0cb7c2ba2760322e.

This commit also adds a few tests that operator precedence in BeanShell behaves the same as Java


nickl-:
Looks good nice work!!

nickl-:
Again reverting commits does not fix problems, it just voids everything that was accomplished.

Find the problem and fix it, do not look for the commit that changed it.

elbosso:
The Fix for the problem is to remove this optimization from hell! please finally merge this pull request!

Issue #607, opened 2021-05-28
JSR223 ScriptEngine does not update globally scoped variables


opeongo:
BeanShell has a JSR223 Script Engine provider.  JSR223 allows scripting languages to plug in to the Java system and interact with each other.  Interactions between scripting engines occurs in the global scope of bindings.  The plugin for BeanShell is able to retrieve values from the global bindings, but it is unable to update existing bindings.  Instead when trying to update a global binding the BeanShell engine will instead shadow the variable in the local (engine) context.

The following sample Java code shows the problem:
```java
ScriptEngineManager manager = new ScriptEngineManager();
ScriptEngine engine = manager.getScriptEngineByName("beanshell");

// Put a value in to the global scope
manager.put("a", 44);

// Check that the value is seen within the BeanShell interpreter
assert(44 == engine.eval("a"));

// Note that this variable is not in the engine scope
assert(engine.get("a") == null);

// Try to update the value
engine.eval("a++");

// Note that the global value did not update
assert(manager.get("a") == 44);

// but now there is a engine scope variable with the expected value
assert(engine.get("a") == 45);
```

This behaviour prevents the BeanShell script engine from having two-way communication with other script engines.  The BeanShell script engine should be fixed to allow updates of globally scoped variables.

PR #608, opened 2021-05-28
Fix the JSR223 script engine

opeongo:
Fix the JSR223 script engine plugin so update variables in the same scope that they are found in.  The previous version would always set variables in the ENGINE_SCOPE, thus causing global scope variables to be shadowed.

Add tests for correctness.

This PR provides a fix for issue #607.

### DIFF
--- a/src/main/java/bsh/engine/ScriptContextEngineView.java
+++ b/src/main/java/bsh/engine/ScriptContextEngineView.java
@@ -111,7 +111,8 @@ public class ScriptContextEngineView implements Map<String, Object> {
 
 
     /**
-     * Set the key, value binding in the ENGINE_SCOPE of the context.
+     * Set the key, value binding in the scope where the key is found, or
+     * in the ENGINE_SCOPE if the key is not found.
      *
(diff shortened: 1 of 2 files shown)


src/main/java/bsh/engine/ScriptContextEngineView.java
```diff
@@ -130,8 +131,19 @@ public Object get(Object key) {
...
-        Object oldValue = context.getAttribute(key, ENGINE_SCOPE);
-        context.setAttribute(key, value, ENGINE_SCOPE);
+       /*
+        * Check the scope for this key.  If not found then set the scope to
+        * ENGINE_SCOPE
+        */
+       int scope = context.getAttributesScope(key);
+       if (scope<0)
+          scope = ENGINE_SCOPE;
+        Object oldValue = context.getAttribute(key);
```
nickl-, line 141: Shouldn't this still have the scope, we know now where to look for it?

```java
Object oldValue = context.getAttribute(key, scope);
```

Issue #508, opened 2018-08-21
failure in run() command corrupts the interpreter


opeongo:
It seems that when the object of the beanshell `run()` command fails some kind of corruption happens to the interpreter.

Here is an example of a script to be run, guaranteed to throw an exception (from fail_me.bsh):
```java fail_me.bsh
throw new Exception("boom");
```
Here is an invoking script that demonstrates the problem (from run_fail.bsh):
```java
try { run("fail_me.bsh"); } catch ( bsh.EvalError e ) { print("First run failed with EvalError"); } 

try { run("fail_me.bsh"); } catch ( Throwable fne ) { print("Second run failed with "+fne); }
```

If I start up an interpreter and source the invoking script 
`source("run_fail.bsh");`
and then follow up with an `exit()` command, the script will produce the following output:
```
C:\github\beanshell>java -cp target\classes bsh.Interpreter
BeanShell 2.1.0-SNAPSHOT - by Pat Niemeyer (pat@pat.net)
bsh % source("run_fail.bsh");
First run failed with EvalError
Second run failed with java.io.FileNotFoundException: C:\github\beanshell\null\fail_me.bsh (The system cannot find the path specified)
--> $0 = null :null
bsh % exit();
// Error: Evaluation Error: Undefined variable or class name while evaluating: b
sh.system.shutdownOnExit : at Line: 10 : in file: /bsh/commands/exit.bsh : bsh .
system .shutdownOnExit

Called from method: exit : at Line: 2 : in file: <unknown file> : exit ( )
bsh %
```

Notice the 'null' that is in the path name given in the exception.  Obviously not a problem during the first invocation.

[run_fail.bsh.txt](https://github.com/beanshell/beanshell/files/2307013/run_fail.bsh.txt)
[fail_me.bsh.txt](https://github.com/beanshell/beanshell/files/2307015/fail_me.bsh.txt)

opeongo:
I had a look in to this a bit more.  It turns out that it has nothing to do with the exception being thrown.  Simply using the run command a second time in the interpreter causes the problem.  Working off the previous `run_fail.bsh` example, just changing the fail_me.bsh file to have:
```java
print("hello");
```
is enough to trigger the problem the second time the run command is invoked.

I tried to dig in to see what was going on and found that the problem showed up at commit de15ce30c79090d4c16478ec139cb41daa044a7d.  Before this commit multiple run commands worked correctly.

[...]

PR #610, opened 2021-10-26
Revert commit de15ce30c79090d4c16478ec139cb41daa044a7d

opeongo:
"Fix override child namespace for namespace chaining. "

This addresses issue #508 and other failures caused by the commit.

### DIFF
--- a/src/test/java/bsh/Namespace_Chaining_Test.java
+++ b/src/test/java/bsh/Namespace_Chaining_Test.java
@@ -50,46 +50,6 @@ public class Namespace_Chaining_Test {
 
 
     @Test
-    public void override_child_namespace() throws Exception {
-        Interpreter root = new Interpreter();
-        Interpreter child = new Interpreter(new StringReader(""), System.out, System.err, false, new NameSpace(root.getNameSpace(), "child"));
-
-        root.eval("int bar=42;");
-        child.eval("int bar=4711;");
-
-        assertEquals(42, root.eval("bar;"));
-        assertEquals(4711, child.eval("bar;"));
-
-        // Method foo declared in parent namespace returns bar
-        root.eval("int foo() { return bar; }");
-
-        assertEquals(42, root.eval("foo();"));
-        assertEquals(4711, child.eval("foo();"));
-    }
-
-    @Test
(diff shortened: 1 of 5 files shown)


nickl-:
Can we hold this one? see #508 

Removing 2 unit tests to make a fix seems extreme.

src/test/java/bsh/Namespace_Chaining_Test.java
```diff
@@ -48,6 +48,45 @@ public void namespace_nesting() throws UtilEvalError {
...
+        assertEquals(42, root.eval("bar;"));
+        assertEquals(4711, child.eval("bar;"));
+        assertEquals(5, child2.eval("bar;"));
+
+        // Method foo declared in parent namespace returns bar
+        root.eval("int foo() { return bar; }");
+
+        assertEquals(42, root.eval("foo();"));
+        assertEquals(42, child.eval("foo();"));
+        assertEquals(42, child2.eval("foo();"));
```
nickl-, line 88: But this is not correct, the method is declared in parent namespace, correct but it is returning the variable `bar`.

Which for `root` is 42, but for `child` has the value of 4711 and for `child2` should be 5.

There is inheritance but gets overwritten if declared or changed.

Lets look at it from another perspective:

```java
class Root {
   int bar = 42;
   int foo() {
      return bar;
   }
}

class Child extends Root {
   int bar = 4711;
}

int baz = new Child().foo();
```

What will the value be for `baz`?

opeongo: The value will be 4711.  

This bug has nothing to do with inheritance, and this reversion will not change your example from working.  The change that this will revert was looking up and using variables in the namespace where the scripted method was being executed in, rather than using the variables from the namespace that the function was declared in.   This behaviour causes the namespace stack to become irrecoverably corrupted and subsequently run scripts to fail so it is a serious bug.  

This bug has been open for almost 4 years, it is a showstopper, and it is blocking other fixes from being merged.   Time to revert and close off this issue.

Additionally I would say that the behaviour was a bit weird anyway.  If there is interest in making a fix to bring this back I think that:
1. there should be a discussion about the proposed behavioural change.
2. all the unit tests should pass.

Issue #613, opened 2021-10-27
Unnecessary casting causing problems with converting list to array 


opeongo:
Commit 32e8be51001be388743dac45376a699c9549c41a introduced casting of arguments to match the types of the method that is being called.  This is essential for the 'magic' of coercing arrays to lists.  It is unnecessary for regular reference types where the new type is a superclass of the existing type, and in fact it causes problems like in this simple case:

```java
list = new ArrayList(Arrays.asList(new Integer[] {1, 2, 3, 4}));
arr = list.toArray(new Integer[0]);
```

This code snippet will result in an ArrayStoreException, because internally the `new Integer[0]` argument will be coerced to `new Object[0]`.

The fix for this is to not coerce regular reference types when the target is a supertype.

nickl-:
#614 RTBC

**Closed: Fixed in PR**

PR #614, opened 2021-10-27
Avoid unnecessary coercion

opeongo:
Fix for issue #613 and add test case


nickl-:
Good primary cause analysis and appropriate fix!

RTBC

Issue #616, opened 2021-10-28
Building with Java 16+


opeongo:
I tried building BeanShell with JDK 16 and ran in to a couple of errors.

1. Jacoco (part of the maven build system) needs to be upgraded to 0.8.7 or the build will fail.
2. One of the methods for opening jar files returns a different exception when passed the name of a jar file that does not exist.  Previously the method returned an unchecked RuntimeException, now it returns an IOException.  This is not a big deal but it does change the error handling mechanism.  One of the tests in the test suite now fails because of this change.  

These are pretty minor problems and easy to fix.  

It seems like a good idea to be able to keep up with the current Java versions so that we don't fall behind too far.  At the moment BeanShell is being built with source=1.8 and target=1.8, and this seems reasonable to keep BeanShell 3.0 with this baseline.  

Going forward it will probably make sense to have a BeanShell 3+ (3.1 or 4.0?) that targets a newer platform version so that we can take advantage of new features, but this is something for the future.

PR #617, opened 2021-10-28
Minor changes to allow building with JDK 16

opeongo:
Update jacoco version to one that is compatible with Java 16
Paper over the change in exception thrown by one of the jdk classes related to jars, and adjust one of the tests to match

Fix for issue #616


slachiewicz (approved):

nickl-:
Does travis support this build yet? Then we should probably include that configuration with this patch.

Issue #615, opened 2021-10-27
Java 16 and strongly encapsulated Java internals causing problems with MethodHandle lookup


opeongo:
Project Jigsaw introduced modules and the access issues that come with them.  Things changed again in Java 16 as strongly encapsulated internals became the default.  Interfaces can be accessed from a module that is exported, but the implementation might be in a module that is not exported.

The following BeanShell example succeeded on Java versions 8 to 15, but fails on Java 16+.  

```java
import javax.xml.parsers.DocumentBuilderFactory;
import org.xml.sax.InputSource;

dbf = DocumentBuilderFactory.newInstance();
dbf.setValidating(false);
dbf.setNamespaceAware(true);
dbf.setFeature("http://xml.org/sax/features/namespaces", false);
dbf.setFeature("http://xml.org/sax/features/validation", false);
dbf.setFeature("http://apache.org/xml/features/nonvalidating/load-dtd-grammar", false);
dbf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
docBuilder = dbf.newDocumentBuilder();
doc = docBuilder.parse(new InputSource(input));
nodes = doc.getElementsByTagName("mapID");
```

This code will fail with `IllegalAccessException` on Java 16+

nickl-:
Damn I'm so not looking forward to battling with these indecision of Oracle, been blissfully unaware of these for the last 2 years. So glad you resolved it already =) 

Also looks like they are rolling out java releases like toilet paper.

Has commit dc0b3b0 made it to upstream yet? (I don't see it in PR)

nickl-:
Found #618 RTBC

**Closed: Resolved in PR**

PR #618, opened 2021-10-28
Fix for issue #615 

opeongo:
Strong encapsulation in Java 16+ causes problems for MethodHandle lookups.
In case there is an AccessException when looking up a method handle this fix will continue searching up the class inheritance hierarchy to find an accessible method.


nickl-:
Ahh here is the #615 patch I missed. Looks good, RTBC.

We should probably include a build in travis if supported.

PR #619, opened 2021-11-03
Fix stack trace in BSHTryStatement

opeongo:
Modify BSHTryStatement to throw the original TargetError that contains the
complete script stack trace, rather than a stack trace that only starts from
the try statement.  Without the complete stack trace it is difficult to
locate where the original error happened.

Also add a test case to TryStatementTest.


nickl-:
Hmm I think this breaks standard conformity, beanshell throws TargetError.  Can't immediately recall but we use this somewhere...hmmm maybe in CLI perhaps?

Agree completely with the problem statement, need to have the stack trace pull through.

Can't we wrap `originalException` and ensure that it's stacktrace is included.

Example:
```java
    throw new TargetError(originalException, this, callstack);
```  

Instead of 

```java
    throw originalException;
```

Which is unwrapped and can't be handled as beanshell specific exception.

Unless I am missing something... Needs work.

opeongo:
@nickl- Yes you are correct that the exception needs to be wrapped in order to complete the callback trace.  I have a fix in my tree for this and will push it when ready.  However this needs #610 to be applied first to fix the problem about namespaces being overridden otherwise the callstack does not build up and no trace gets written out (another side effect from de15ce30c79090d4c16478ec139cb41daa044a7d).

nickl-:
I think the reason why the message might be getting lost is here:

https://github.com/beanshell/beanshell/blob/680765207eca5c190d55009cc1f4a589365070eb/src/main/java/bsh/TargetError.java#L56-L59

Perhaps we should append  `"TargetError:" + t.getMessage()` or just pass the Throwable's message alone. I think I prefer the first option...

Either way this seems like the better spot to apply the fix.

nickl-:
Still think the issue is actually in `TargetError` obfuscating the message.

Needs build errors/checks resolved.

nickl-:
Got the error located at `try {` nonsense again at #682 a fix would be awesome

jimjag:
Closing during a backlog triage pass: this is blocked on a separate unmerged PR (#610) and has drifted from master with unresolved build errors. Reopen if this is still relevant.

Issue #620, opened 2021-11-05
Not correctly selecting VarArgs methods


opeongo:
I had a method call on an overridden method blow up with an index out of range exception in the BeanShell code. It seems that the overridden method that was selected had fewer parameters than the arguments that I was passing in.  Surely this should not have happened that the selected method did not match the caller.  After some digging around I came across a two problems.  The code is getting a bit hard to follow, so to keep things as simple as possible I will unpack these problems one by one and propose fixes one by one. 

The first problem I will look at was exposed by a different problem, so I will fix this first before fixing the other problem that will hide this one again.  It looks like the mechanism for selecting the most specific overridden method is not working properly for VarArgs methods.  The rules in[ JLS 15.12.2](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2) have the order

1. check fixed-arity methods with strict context
2. check fixed-arity methods with loose context (boxing, narrowing)
3. check variable-arity methods with loose context.

As soon as one of these match then the search is over.  However in the current code base the variable arity and fixed arity methods get lumped together at the start of the algorithm and tested simultaneously in steps 1 and 2.

Consider this sample code:
```java
    // for testing VarArgs and most specific method selection
    public String method2(String str) {
        // most specific for method2("Hello")
        return "one";
    }
    public String method2(String str, String str1) {
        // most specific for method2("Hello", "World")
        return "two";
    }
    public String method2(String str, String... str2) {
        // most specific for method2("Hello", "World", "This", "is", "me")
        return "three";
    }
```

The current master BeanShell selects the following:
```java
m.method2("Hello") returns "one"
m.method2("Hello", "World") return "three"
m.method2("Hello", "World", "This", "is", "me") returns "three";
```

I had a look at the code and it seems like there are two separate mechanisms for handling varargs.  The first is started in [Reflect.findMostSpecificInvocable](https://github.com/beanshell/beanshell/blob/cb552a4d08c279680d6463b61e2f034fe12a590b/src/main/java/bsh/Reflect.java#L431).  This one looks like it is on the right track but it does not look like it was fully implemented.  

[...]

PR #621, opened 2021-11-08
Better handling of VarArgs methods and constructors.

opeongo:
Fix for issue #620.
Closer following of JLS 15.12.2 method resolution rules.
Propagate VarArgs flag to generated methods.
Better marshalling of arguments in varargs methods.
Update test cases.

This touched a few more files than I expected but it turned out that there were a few more loose ends.


nickl-:
Looks good...RTBC

PR #624, opened 2021-11-13
fixes #623

### LINKED ISSUE #623: vulnerable against Trojan Source Attack
The atttack is outlined here: https://arxiv.org/abs/2111.00169

If one copies one of the [examples](https://github.com/nickboucher/trojan-source) into the Console, this happens:
image

The user does not see that there is a possible problem here.
There are many different mitigations for this to be seen in the wild - GitHub shows a message, some projects (intellij idea for example) simply show the problematic control characters as pseudo-glyphs...
I think BeanShell should also do something about it!

elbosso:
This proposal exchanges the text component used in the JConsole  for one that does have syntax highlighting and also mitigates the trojan source attack by visualizing the effective code:
image


nickl-:
See #623 changing the simple text editor is overkill and unnecessary. What is important is that Beanshell does not interpret the vulnerability and execute something which the user might not expect. Throwing an exception and informing the user about the issue should suffice.

This protects everything and not just code from the text editor.

Will leave time to discuss but am leaning towards closing this: won't implement.

nickl-:
Note: We also need travis and codecov fixed, before we can review and merge PRs. I notice it's stopped running here too. =(

nickl-:
Closed: won't implement. As per #623

PR #648, opened 2022-08-02
Block namespace methods fix.

### LINKED ISSUE #508: failure in run() command corrupts the interpreter

nickl-:
Fixes #508
Override namespace looses its local namespace  flag, causing duplicate blocked methods to fail on subsequent calls.
Added unit tests to prove fixes.

### DIFF
--- a/src/main/java/bsh/BshMethod.java
+++ b/src/main/java/bsh/BshMethod.java
@@ -343,11 +343,9 @@ public class BshMethod implements Serializable {
         if ( overrideNameSpace )
             localNameSpace = callstack.top();
         else
-        {
             localNameSpace = new NameSpace( declaringNameSpace, name );
-            localNameSpace.isMethod = true;
-        }
-        // should we do this for both cases above?
+
+        localNameSpace.isMethod = true;
         localNameSpace.setNode( callerInfo );
 
         /*
--- a/src/test/resources/test-scripts/eval.bsh
+++ b/src/test/resources/test-scripts/eval.bsh
@@ -36,4 +36,15 @@ assert(b==6); // sanity check
 assert(s.a==5);
 assert(a==6);
 
+// block namespace calling method twice #508
+y = 22;
+try { eval("yval=y;"); }
+catch (e) { assert(false); }
+assert(yval==y);
+yval=0;
+try { eval("yval=y;"); }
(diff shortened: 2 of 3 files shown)


opeongo:
Wrong fix.  With the caching of namespaces this will cause problems when a cached namepace gets marked as isMethod and then gets reused in a non-method context.

nickl-:
> Wrong fix. With the caching of namespaces this will cause problems when a cached namepace gets marked as isMethod and then gets reused in a non-method context.

You need to elaborate please, how does a non method namespace get used as a method's namespace?

PR #651, opened 2022-11-01
Fix small NPE on Android

mahesh-hegde:
Fixes a small NPE on Android, which happens because Security.getProperty call returns null there.


nickl-:
Thank you for this!

For some reason the CI didn't run so no way to test the regression, will try get that going.

mahesh-hegde:
Seems to be concurrency related error. Not sure what's causing it.

I didn't test locally, because some tests fail on my machine.

No problem if this cannot be merged. Just thought someone may want to run it on Android in future.

nickl-:
There may be something wrong with the CI scripts not sure why it stopped running automatically or why it is failing all the open PRs. Will need to make time and come through this...

PR #653, opened 2022-12-01
Sync with old patches

jimjag:


nickl-:
I am not following, what are we trying to accomplish here?

nickl-:
@jimjag Looks like you're on a big mission, if you need a hand with anything shout.

Issue #667, opened 2022-12-10
Documented magic invoke method fixed


nickl-:
## Synopsis 
Usage of a "catch all" invoke method is documented in several places, yet its implementation is arguably inconsistent. For the most part it appears to still be working as intended, but at closer inspection much is left to be desired. It appears to work for class loaded interfaces and anonymous classes, but it does not work when referencing generated classes or interfaces.

Example invoke method:
```java
invoke( name, args ) { print("Method: "+name+" invoked!"); }
```
## Documentation
There are two references in documentation to its use for interfaces and another for commands:
1.  [Implementing Interfaces](https://beanshell.github.io/manual/bshmanual.html#Implementing_Interfaces)
2. ["Dummy" Adapters and Incomplete Interfaces](https://beanshell.github.io/manual/bshmanual.html#%22Dummy%22_Adapters_and_Incomplete_Interfaces)
3. [User Defined Commands with invoke()](https://beanshell.github.io/manual/bshmanual.html#User_Defined_Commands_with_invoke())
Which is different to the [compiled commands](https://beanshell.github.io/manual/bshmanual.html#Compiled_Commands) invoke method.

## Inconsistencies
On the one hand it does not appear to be a "catch all" on objects per se, invoke will not be called for unknown method but only for unimplemented interface methods. These methods are reported by reflection as being abstract but it is not clear if this is also intended for unimplemented abstract methods from a parent abstract class, **whether this was implemented in BeanShell 2.x needs to be verified**. On the other hand if declaring an invoke function (as per user defined commands) it will be called for any unknown function or "Command not found" exception, being truly a "catch all". Would we not want to expect the same functionality across the board? 

Then when invoke is used for writing compiled command classes, it has a completely different signature, accepting namespace and a call stack instead of name and arguments. This may have been regarded as a special use case but not when scripting commands is mentioned in the same context. Whether this is an inconsistency that needs to be addressed in implementation or if an update to the documentation separating "catch all" invoke from compiled commands will suffice to define a clear distinction of use case, is open for debate.

### Working

Using invoke to catch all functions:
```java
bsh % abc();
// Error: Evaluation Error: Command not found: abc() : at Line: 1 : in file: <unknown file> : abc ( )

bsh % invoke( name, args ) { print("Method: "+name+" invoked!"); }
--> void
bsh % abc();
Method: abc invoked!
--> void
bsh % this.abc();
Method: abc invoked!
--> void
```
[...]

PR #668, opened 2022-12-10
Add invoke command to This.invokeMethod.

### LINKED ISSUE #115: Fix to This For Compiled Commands
The following is a fix to the This class so that it
corresponds to the way that the Name class will first
attempt to getMethod\(\) from the namespace and then try
to do a getCommand\(\) on the namespace. The reason for
this patch is that when a namespace object had imported
commands from a location where some commands are
compiled java code commands, the commands could not be
accessed when not inside that particular namespace.

e.g.

aNamespace.someCompiledJavaCodeCommand\("foo"\);

This case would not work without the following patch:

/\*
invokeMethod\(\) here is generally used by outside code
to callback
into the bsh interpreter. e.g. when we are acting as
an interface
for a scripted listener, etc.  In this case there is
no real call stack
so we make a default one starting with the special
JAVACODE namespace
and our namespace as the next.
\*/
public Object invokeMethod\( 
String methodName, Object \[\] args, 
Interpreter interpreter, CallStack callstack,
SimpleNode callerInfo, 
boolean declaredOnly  \) 
throws EvalError
\{
/\*
Wrap nulls.
This is a bit of a cludge to address a deficiency in
the class
generator whereby it does not wrap nulls on method
delegate.  See
Class Generator.java.  If we fix that then we can
remove this.
\(just have to generate the code there.\)
\*/
if \( args \!= null \)
\{
Object \[\] oa = new Object \[args.length\];
for\(int i=0; i&lt;args.length; i++\)
oa\[i\] = \( args\[i\] == null ? Primitive.NULL : args\[i\] \);
args = oa;
\}

if \( interpreter == null \)
interpreter = declaringInterpreter;
if \( callstack == null \)
callstack = new CallStack\( namespace \);
if \( callerInfo == null \)
callerInfo = SimpleNode.JAVACODE;

// Find the bsh method
Class \[\] types = Types.getTypes\( args \);
BshMethod bshMethod = null;
try \{
bshMethod = namespace.getMethod\( methodName, types,
declaredOnly \);
\} catch \( UtilEvalError e \) \{
// leave null
\}

if \( bshMethod \!= null \)
return bshMethod.invoke\( args, interpreter,
callstack, callerInfo \);

/\*
No scripted method of that name.
Implement the required part of the Object protocol:
public int hashCode\(\);
public boolean equals\(java.lang.Object\);
public java.lang.String toString\(\);
if these were not handled by scripted methods we
must provide
a default impl.
\*/
// a default toString\(\) that shows the interfaces we
implement
if \( methodName.equals\("toString" \) \)
return toString\(\);

// a default hashCode\(\)
if \( methodName.equals\("hashCode" \) \)
return new Integer\(this.hashCode\(\)\);

// a default equals\(\) testing for equality with the
This reference
if \( methodName.equals\("equals" \) \) \{
Object obj = args\[0\];
return new Boolean\( this == obj \);
\}

// Look for a default invoke\(\) handler method in the
namespace
// Note: this code duplicates that in NameSpace
getCommand\(\)
// is that ok?
try \{
bshMethod = namespace.getMethod\( 
"invoke", new Class \[\] \{ null, null \} \);
\} catch \( UtilEvalError e \) \{ /\*leave null\*/ \}

// Call script "invoke\( String methodName, Object \[\]
args \);
if \( bshMethod \!= null \)
return bshMethod.invoke\( new Object \[\] \{ methodName,
args \}, 
interpreter, callstack, callerInfo \);
// We will do one last desperate attempt. We will see
if the method being
//  called on us is for a compiled command
else \{
Class \[\] argTypes = Types.getTypes\(args\);
Object commandObject = null;
try \{
commandObject =
namespace.getCommand\(methodName,argTypes,interpreter\);
\} catch \(UtilEvalError e\) \{
// Do nothing, just drop out and allow the method
to exit normally.
\}

if \(commandObject \!= null && commandObject
instanceof Class\) \{
try \{
return
Reflect.invokeCompiledCommand\(\(Class\)commandObject,args,interpreter,callstack\);
\} catch \(UtilEvalError e\) \{
// Do nothing, just drop out and let the method
exit normally.
\}
\}
\}

throw new EvalError\("Method " + 
StringUtil.methodString\( methodName, types \) +
" not found in bsh scripted object: "+
namespace.getName\(\), 
callerInfo, callstack \);
\}

Reported by: *anonymous

Original Ticket: "beanshell/patches/9":https://sourceforge.net/p/beanshell/patches/9

nickl-:
Refactored invoke command from `Name` to `Namespace`
Added invoke command to `This.invokeMethod` fixes #115
Removed duplicate invoke method implementation as per #667
Added complete coverage unit tests for `This.invokeMethod`
Fixed deprecations on `TestBshScriptEngine`

### DIFF
--- a/src/conf/version.properties
+++ b/src/conf/version.properties
@@ -14,6 +14,6 @@
 #
 #Project version number - populated by maven
 #Project build number - incremented by beanshell-maven-plugin
-#Sat Dec 10 17:26:46 SAST 2022
-build=1038
+#Sat Dec 10 23:18:18 SAST 2022
+build=1094
 release=${project.version}
--- a/src/conf/version.properties
+++ b/src/conf/version.properties
@@ -14,6 +14,6 @@
 #
 #Project version number - populated by maven
 #Project build number - incremented by beanshell-maven-plugin
-#Thu Dec 08 16:32:51 SAST 2022
-build=936
+#Sat Dec 10 17:26:46 SAST 2022
+build=1038
 release=${project.version}
--- a/src/conf/version.properties
+++ b/src/conf/version.properties
@@ -14,6 +14,6 @@
 #
 #Project version number - populated by maven
 #Project build number - incremented by beanshell-maven-plugin
-#Sat Dec 10 23:18:18 SAST 2022
-build=1094
+#Sun Dec 11 05:00:12 SAST 2022
+build=1169
 release=${project.version}
--- a/src/main/java/bsh/BSHAllocationExpression.java
+++ b/src/main/java/bsh/BSHAllocationExpression.java
@@ -82,14 +82,10 @@ class BSHAllocationExpression extends SimpleNode
         if ( args == null)
             throw new EvalError( "Null args in new.", this, callstack );
 
-        // Look for scripted class object
+        // Lookup class
         Object obj = nameNode.toObject(
-            callstack, interpreter, false/* force class*/ );
+            callstack, interpreter, true /*force class*/ );
 
-        // Try regular class
-
-        obj = nameNode.toObject(
-            callstack, interpreter, true/*force class*/ );
 
         Class<?> type = null;
         if ( obj instanceof ClassIdentifier )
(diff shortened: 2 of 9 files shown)


nickl-:
Fix generated interface proxy recursion loop.

When a generated interface is used to create a dynamic proxy instance, access to its NS field gets obscured.
Removed a duplicate class lookup in `BSHAllocationExpression` `eval` method, which appears to have been there forever.
Fixed up some missing Class generics and added a few more debug messages.

Issue #109, opened 2003-11-07
Allow access to exit return code


sourceforge-issue-exporter:
It would be useful sometimes to have access to the exec() return code.

Here is my amended version of the exec() code:

int // NEW
```java
exec( String arg ) {
    this.proc = Runtime.getRuntime().exec(arg);
    this.din = new DataInputStream(proc.getInputStream() );
    while( (line=din.readLine()) != null )
        print(line);
    return this.proc.exitValue(); // NEW
}
```

Reported by: *anonymous

Original Ticket: "beanshell/patches/3":https://sourceforge.net/p/beanshell/patches/3

nickl-:
This is handy but perhaps early access to [the documentation for Process](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Process.html)  itself is more useful.
 
As also mentioned in #273 and as per the above documentation page we should change to ProcessBuilder

> As of 1.5, [ProcessBuilder.start()](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/ProcessBuilder.html#start()) is the preferred way to create a Process.

Yet keep backwards compatibility...

nickl-:
Replaced by #683

PR #670, opened 2022-12-13
low memory unit tests

nickl-:
optimised unit test memory usage, fixed deprecation fixed two incomplete tests, added coverage for BshClassPath return exit status in exec command as per #109
fix deprecated File.toURL in addClassPath command
Cleaned up BshClassPath with minor refactoring
Unsubscribe BSHType from class loader changed broadcasts


nickl-:
Duplicate of #669 which stopped doing the automations.
Just need to look into the synchronicity of BshClasspath again, think I figure lift out...

PR #671, opened 2022-12-14
Fix the problem with operator precedence (#590)

opeongo:
- set up productions in reverse order of precedence
- split productions that contain operators of different precedence
- join productions that have operators of similar precedence
- use tokens instead of strings
- add test case
- minor cleanup

This should fix the problem with 1+2*3=9 as noted in #590 and probably elsewhere.  Simply reverting the patch that broke operator precedence wasn't acceptable, and with no one else stepping up I went ahead and redid this.  So basically this RP is almost the same as before but fixes mod/power precedence.  I retained the reduction of productions in the unary expression and elimination of one lookahead from the cast expression.  Tokens are used instead of strings because a one point this was the strong recommendation for better performance from the javacc documentation.  The test case is a work in progress.

### DIFF
--- a/src/main/jjtree/bsh.jjt
+++ b/src/main/jjtree/bsh.jjt
@@ -775,62 +775,164 @@ int AssignmentOperator() :
     }
 }
 
+/*
+ * Here follows the productions for Java expressions.
+ * JavaCC implements operator precedence using the cascade of
+ * productions.  It is important to maintain these as separate
+ * independent productions in the reverse sequence of the order of
+ * precedence. Do not flatten these in to fewer productions in an
+ * attempt to make the parser go faster because:
+ * - it will lose all the BEMDAS/PEMDAS order of operations.
+ * - it probably wont go faster.  Don't forget JavaCC will emit java
+ * code that still has all of the conditional tests.  All flattening
+ * will do is rearrange them in to a different order.  The java code 
+ * will be compiled, jitted and inlined by c2 so you wont even know
+ * what the performance impact is unless you do micro benchmarks.
+ */
 void ConditionalExpression() :
-{ Token t; }
+{ }
 {
-  RelationalExpression() (  "?" ConditionalExpression() ":" ConditionalExpression() #TernaryExpression(3)
-  |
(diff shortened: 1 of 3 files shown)


src/main/jjtree/bsh.jjt
```diff
@@ -775,62 +775,164 @@ int AssignmentOperator() :
...
+ * independent productions in the reverse sequence of the order of
+ * precedence. Do not flatten these in to fewer productions in an
+ * attempt to make the parser go faster because:
+ * - it will lose all the BEMDAS/PEMDAS order of operations.
+ * - it probably wont go faster.  Don't forget JavaCC will emit java
+ * code that still has all of the conditional tests.  All flattening
+ * will do is rearrange them in to a different order.  The java code 
+ * will be compiled, jitted and inlined by c2 so you wont even know
+ * what the performance impact is unless you do micro benchmarks.
+ */
```
nickl-, line 791: This comment should only shed light on what the code does, if it is unclear otherwise. It should not be the motivation for the change, that is what the PR and commit messages are for.

nickl- (changes requested):
I was also thinking the easiest way to resolve this is to build the hierarchy in the grammar. This does not mean that there are no other way to deal with this.

The comment will not be useful 10 years from now.

nickl-:
Also have you considered addressing the [other outstanding parser issues](https://github.com/beanshell/beanshell/issues?q=is%3Aissue+is%3Aopen+label%3AParsing) while at this?

nickl-:
Please also add unit test for:
```java
true && false ? true : true && false ? true : false;
```

Just to be sure.

opeongo:
> Please also add unit test for:

done

opeongo:
> the easiest way to resolve this

Are there still items that need to be resolved on this issue?  I think at this point you will have to put your own code forward.

nickl-:
> > the easiest way to resolve this
> 
> Are there still items that need to be resolved on this issue? I think at this point you will have to put your own code forward.

Looks good, RTBC

nickl-:
This is RTBC - ready to be committed (merged) 

Are you waiting for me?

Issue #659, opened 2022-12-02
Parameter/variable name lookup confusion


opeongo:
As I was putting together a test script for #655 I noticed a problem with parameter/variable name lookups.  The following script fails at line 23 with the error message:
 Attempt to resolve method: add() on undefined variable or class name: c.ylds : at Line: 23 

```java
test1(String c) {  //change String c to String d and it works!
   return this;
}

container() {
   System.err.println("allocate container");
   ylds = new ArrayList();
   work() {
      System.err.println("working");
   }
   return this;
}

c = null;
for (int i=0; i<2; i++) {
   for (int j=0; j<10; j++) {
      if (j%5 == 0) {
         if (c != null) {
            c.work();
         }
         c = container();
      }
      c.ylds.add(test1("hello"));  // line 23
   }
}
```

However, if I change the formal parameter name in the `test1` method from `c` to `d` the script will work successfully.

What am I missing here?

nickl-:
The script calls line 23 every iteration but only ` c = container();` on the 5th. So `c.ylds` is undefined for 0 to 4.

The parameter name of the `test1` method should have no effect.

opeongo:
Your mental interpreter did not execute this correctly.  If you try running it you will see that the container is created when `j == 0` (`0%5 == 0` by the way).  If `c` was null there would have been a different error.

I agree that the name of the parameter to the `test1` method should not have any effect,  But in this test case it appears to do so,  which indicates some kind of lookup mistake.

opeongo:
Here is a revised minimal test script that isolates the problem a bit better.  No loops, no if's to distract the focus.

```java
test1(String c) {  // Change String c to String d and it works
   return this;
}

container() {
   System.err.println("allocate container");
   ylds = new ArrayList();
   return this;
}

Object c = null;
{
   {
      c = container();
   }
   // c = container();  // uncomment this line and it works correctly
   c.ylds.add(test1("hello"));  // Line 17, Error here.  Comment out and it will work
}
c.ylds.add(test1("hello"));  // Line 19, This works
```
Something happens when allocations are made within block namespaces.  Note that as the comments state that this script can be modified to work correctly by either

1. renaming the parameter to the `test1` method, or
2. instantiating the container object outside of the inner block, or
3. Commenting out line 17, and the script will succeed at line 19.

You don't have to change more than one condition.  Any of these three changes independently will avoid the error.

[...]

Issue #655, opened 2022-12-02
Variables disappearing in long running scripts


opeongo:
I have noticed a weird behaviour that occurs in long running scripts that have lots of memory churn.  What happens is that a loop will work successfully for several 10's of thousands of iterations, and then it will error out that an operator is not appropriate.  The line of code is one that was successfully run previously, but at this iteration it simply broke.

This is a tricky bug to isolate and reproduce because it is intermittent and non-deterministic.

Here is an example:

bsh.EvalError: Sourced file: xxxxxxxxxxx.bsh : Unary operation "++" inappropriate for object : at Line: 976 : in file: xxxxxxxxxxxxxx.bsh : ) { 

The line in question in:
```java
   for (int k=0; k<inputs.length; k++) {  // line 976
```
This loop was repeating after 15 minutes of processing when the error occurred.

Here is another example:
bsh.EvalError: Sourced file: xxxxxxxxxx.bsh : Not an array or List type : at Line: 380 : in file: xxxxxxxxxxxxxx.bsh : [ i ] 

The code for this line is:
```java
Object yield(int year, double ba,double qmd, double cag, double cbg, double cinv, double vlog,
             double[] oh, double[] df, double[] oc, double[] ce, double[] ww, double[] pn) {

   for (int i=0; i<4; i++) {
      oh[i] *= (1-0.11);  // line 380
      df[i] *= (1-0.07);
      oc[i] *= (1-0.08);
      ce[i] *= (1-0.11);
      ww[i] *= (1-0.1);
      pn[i] *= (1-0.07);
   }
```
There is only one call site for this scripted object and it calls the object ten's of thousands of times successfully, and then boom it fails.

When I look at what is going on with the debugger it turns out that the lookup of the ```k``` or ```oh``` variables is returning void, where in previous iterations the object was found correctly.  I suspect that this has something to do with the [ReferenceCache](https://github.com/beanshell/beanshell/blob/master/src/main/java/bsh/util/ReferenceCache.java) that uses weak references to [BlockNameSpace](https://github.com/beanshell/beanshell/blob/a36bd52b5037ae9b442a077888a5e4874ac03ce0/src/main/java/bsh/BSHBlock.java#L43) objects and they are getting cleared too soon, but this is guessing.  I am having a difficult time confirming this because of the nondeterminism and length of time it takes for the error to show up.  By the time that the exception happens the 'cause' has already happened elsewhere.

Here is another one, I suppose that this is related:
[...]

PR #672, opened 2022-12-16
Fix for #655 and #659

opeongo:
Here is a patch that works for me for fixing #655 and also #659.

There isn't a lot in this patch, really the significant bits are only two relatively small changes.  That plus wrapping try-finally around a few method calls, reformatting the code to not make my eyes bleed (well, you might not like my style any better), fixing a data race issue, and adjusting the tests to the newly expected results.

I expect that these changes will be controversial and probably not acceptable, but I am putting this out here to show you what is working for me.  If leads to a productive discussion that's great.  If you don't like this approach that's ok too, at least I have shown you something that makes a couple of problems go away.

### Fixing scripted object/variable confusion (#659)
The first change I have already discussed and that is in the `Name` class [to sure that invokeMethod does not set overrideChild flag to true](https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/main/java/bsh/Name.java#L910).  As I discussed elsewhere my analysis leads me to believe (I'm hedging here because I am open to be proven wrong) that method invocations always require that a new namespace be allocated with the declaring namespace as the parent, and this gets swapped on to the top of the callstack.  The body of the scripted object will be run using this new namespace.  When the methods returns `this` to the caller it will be backed by this fresh namespace.  For that reason the `overrideChild` flag must always be false.  I do not think that this is too controversial.

This change will alter the name resolution semantics for closures back to what they were in BeanShell 2.*.   These semantics got changed with de15ce30c.   In previous beanshell versions scripted objects resolved names from with the scripted object first and then through the declaring namespace parent.  The scope of name resolution is then fixed for all all objects created by the method.  de15ce30c changed this so that name resolution was via the top of current call stack. There are two problems here:
1. since names are set in the namespace on the top of the callstack they are not isolated.  One scripted object will clobber the values of the previous object.
2. since the namespace at the top of the current callstack is used for name and value resolution this means that the scripted objects will potentially lookup different variables and values when called from contexts.  

There is nothing illegal about the second point, but it is a significant change from beanshell 2.* and I think it is worth a discussion before breaking backward compatibility.  My opinion is that a) I liked the original way much better because it encapsulates values better; and b) I don't like breaking compatibility.  This new way opens up too many possibilities for variables to be overridden in unexpected and surprising ways.  It's like leaving a loaded gun on the table.  If you need to use different values in different contexts then pass them in as parameters, easy-peasy. Don't just automagically absorb them from enclosing namespaces, this new way is too surprising and difficult to reason about.  It allows third party objects to change how your objects work by setting local variables that will override the closures.  If beanshell were a popular scripting language this would be seen as a security hole not a feature.

So this change is [basically a one-liner](https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/main/java/bsh/Name.java#L910), plus changing the test cases to reflect that variable resolution is via the declaring namespace, not the top of the current callstack

### Disappearing variables (#655)
This problem is that in a long running data processing scripts variables would mysteriously become undefined, when for they previous x00,000 iterations they had worked correctly (read more about this in #655).  My analysis of the problem is described below, and the solution I am using is basically [a three-liner](https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/main/java/bsh/BSHBlock.java#L143).   My solution works for me, and I think my analysis is correct but I am going to hedge here again and say I am open to being proven wrong.

Class BSHBlock is using the ReferenceCache to keep hold of and recycle NameSpace objects.  The current top namespace is the key that is used to access the subordinate namespace's in the cache.   Instead of allocating a new NameSpace object as required, and then allowing it to be disposed of by the GC when it isn't used anymore, the ReferenceCache will create a new NameSpace when there is a miss in the cache, and afterwards will recycle the previously initialized object.  Every time the NameSpace object is finished being used in BSHBlock some of the internal state of the NameSpace is reinitialized.  The cache uses WeakReferences so cache entries can be reaped by the GC, but presumably the hot ones will stay around to a while.  This method depends on GC performance patterns that have no guarantees.

I suppose the ReferenceCache is being used to improve upon the allocation/gc costs of using `new` every time.  Do I understand this right?

If I am correct in how I understand this, then the ReferenceCache will return the same subordinate namespace object for a given namespace.  If three scripted object are manufactured within the same block, would the cache not return the same subordinate namespace for use in each of them?  This would be OK if each block execution were independent, but scripted objects might return `this`, and if recycling namespaces with the cache then wouldn't all three `this` object be backed by the same NameSpace object?  Please tell me I am wrong about this.  Anyway the empirical evidence is that this change made the disappearing variable problem go away.

As I said, my fix is [a simple three-liner](https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/main/java/bsh/BSHBlock.java#L143).  Just instantiate a new NameSpace to use instead of getting a recycled one from the cache.

### Other changes
1. Make the processing of enum constants thread safe.  Previously a static List was used that potentially could have collisions from multiple threads at the same time.
2. Format and and add comments to BSHBlock.  I wanted to make sure that I understood what was going on, so I spent a bit of time on staring at the code and my OCD took over.
3. Add a few try-finally wrappers to make sure that added namespaces are removed off the stack if an exception happens.

### DIFF
--- a/src/main/java/bsh/BshMethod.java
+++ b/src/main/java/bsh/BshMethod.java
@@ -345,6 +345,11 @@ public class BshMethod implements Serializable {
         else
             localNameSpace = new NameSpace( declaringNameSpace, name );
 
+        // This look dubious because in the case of overrideNameSpace
+        // because will clobber the previous value.  OverrideNameSpace==true
+        // is only used in this codebase by PreparsedScripts and there it is
+        // safe but any other usage might cause problems.  This should be
+        // cleaned up.It was changed as part of #508 but that fix was wrong.
         localNameSpace.isMethod = true;
         localNameSpace.setNode( callerInfo );
 
@@ -447,16 +452,19 @@ public class BshMethod implements Serializable {
         if ( !overrideNameSpace )
             callstack.push( localNameSpace );
 
-        // Invoke the block, overriding namespace with localNameSpace
-        Object ret = methodBody.eval(
-            callstack, interpreter, true/*override*/ );
 
-        // save the callstack including the called method, just for error mess
-        CallStack returnStack = callstack.copy();
-
-        // Get back to caller namespace
-        if ( !overrideNameSpace )
-            callstack.pop();
+        // Invoke the block
+        Object ret = null;
+        CallStack returnStack = null;
+        try {
+           ret = methodBody.eval(callstack, interpreter, true/*override*/ );
+           // save the callstack including the called method, just for error mess
+           returnStack = callstack.copy();
+        } finally {
+           // Get back to caller namespace
+           if ( !overrideNameSpace )
+              callstack.pop();
+        }
 
         ReturnControl retControl = null;
         if ( ret instanceof ReturnControl ) {
--- a/src/test/resources/test-scripts/scripted_object.bsh
+++ b/src/test/resources/test-scripts/scripted_object.bsh
@@ -1,0 +1,39 @@
+#!/bin/java bsh.Interpreter
+
+source("TestHarness.bsh");
+
+/*
+ * Test that variable lookups in scripted objects work correctly
+ * Related to #659
+ */
+
+test1(String c) {
+   return this;
+}
+
+container() {
+   ylds = new ArrayList();
+   return this;
+}
+
+Object c = null;
+{
+   {
+      c = container();
+   }
+   c.ylds.add(test1("hello"));
+}
+assert("hello".equals(c.ylds.get(0).c));
+Object d = null;
+{
+   {
+      d = container();
+   }
+   d.ylds.add(test1("goodbye"));
+}
+assert(c.ylds.size() == 1);
+assert("hello".equals(c.ylds.get(0).c));
+assert(d.ylds.size() == 1);
+assert("goodbye".equals(d.ylds.get(0).c));
+
+complete();
--- a/src/test/resources/test-scripts/scripted_object.bsh
+++ b/src/test/resources/test-scripts/scripted_object.bsh
@@ -19,6 +19,15 @@ container() {
 Object c = null;
 {
    {
+      Object c = container();
+      c.ylds.add(test1("hello"));
+      assert("hello".equals(c.ylds.get(0).c));
+   }
+   assert(c == null);
+}
+
+{
+   {
       c = container();
    }
    c.ylds.add(test1("hello"));
--- a/src/main/java/bsh/BSHAllocationExpression.java
+++ b/src/main/java/bsh/BSHAllocationExpression.java
@@ -212,8 +212,11 @@ class BSHAllocationExpression extends SimpleNode
         NameSpace namespace = callstack.top();
         NameSpace local = new NameSpace(namespace, "AnonymousBlock");
         callstack.push(local);
-        body.eval( callstack, interpreter, true/*overrideNamespace*/ );
-        callstack.pop();
+        try {
+           body.eval( callstack, interpreter, true/*overrideNamespace*/ );
+        } finally {
+           callstack.pop();
+        }
         // statical import fields from the interface so that code inside
         // can refer to the fields directly (e.g. HEIGHT)
         local.importStatic( type );
--- a/src/main/java/bsh/BSHBlock.java
+++ b/src/main/java/bsh/BSHBlock.java
@@ -31,21 +31,32 @@ package bsh;
 import java.util.ArrayList;
 import java.util.List;
 
-import bsh.util.ReferenceCache;
-import bsh.util.ReferenceCache.Type;
-
-class BSHBlock extends SimpleNode
-{
+/**
+ * A node reprresenting a code block
+ */
+class BSHBlock extends SimpleNode {
+    /**
+     * The block has a synchronized caluse and requires synchronization
+     */
     public boolean isSynchronized = false;
+
+    /**
+     * This block has a static modifier, so to be used as a static
+     *  initialization block within a class.
+     */
     public boolean isStatic = false;
-    private boolean hasClassDeclaration = false, isFirst = true;
-    private static final List<Node> enumBlocks = new ArrayList<>();
-    private static final ReferenceCache<NameSpace,NameSpace> blockspaces
-        = new ReferenceCache<NameSpace, NameSpace>(Type.Weak, Type.Weak, 4000) {
-            protected NameSpace create(NameSpace key) {
-                return new BlockNameSpace(key);
-            }
-    };
+
+    /**
+     * A flag to immplement a minor optimization of skipping
+     * class declarations when we know there are none.
+     */
+    private boolean hasClassDeclaration = false;
+
+    /**
+     * A flag to immplement a minor optimization of only checking
+     * for class declarations the first time through.
+     */
+    private boolean isFirst = true;
 
     BSHBlock(int id) { super(id); }
 
@@ -55,15 +66,38 @@ class BSHBlock extends SimpleNode
         return eval( callstack, interpreter, false );
     }
 
-    /**
-        @param overrideNamespace if set to true the block will be executed
-        in the current namespace (not a subordinate one).
-        <p>
-        If true *no* new BlockNamespace will be swapped onto the stack and
-        the eval will happen in the current
-        top namespace.  This is used by BshMethod, TryStatement, etc.
-        which must intialize the block first and also for those that perform
-        multiple passes in the same block.
+   /**
+    * Evaluate this block using the top of the callstack to resolve names
+    * and in the context of the given interpreter object.
+    * <p>
+    * In the normal course of events each eval will allocate a new namespace
+    * object using the top of the stack as the parent, and then
+    * this new namespace will be swapped on to the top of the namespace
+    * stack.  During block evaluations this new namespace will be used
+    * for local variables.  When the block is finished the new namespace
+    * is released and then old namespace is restored to the top of the stack.
+    * <p>
+    * There are some situations where the caller has already set up the
+    * namespace ahead of time and does not want another namespace,
+    * for example:
+    * <ul>
+    * <li>BshMethod.invokeImpl()
+    * <br>The method namespace has already been set up containing
+    * the formal parameters.
+    * <li>BSHAllocationExpression.constructWithInterfaceBody()
+    * <br>The caller sets up a namespace the same as if it would be
+    * done using overrideChild=false.  I cannot see the purpose for this.
+    * <li>BSHEnumConstant.eval()
+    * <br>The enum constants should be set within the current namespace.
+    * </ul>
+    * In these situations where the overrideChild flag has been set to true
+    * then use the namespace at top of the callstack.
+    * @param callstack the stack of namespace chains ued to resolve names
+    * @param interpreter the interpreter object used for evaluation
+    * @param overrideNamespace if true the block will be executed
+    * in the current namespace. If false then create an new namespace
+    * and put it at the top of the callstack, restoring the top of the
+    * callstack when done.
     */
     public Object eval(
         CallStack callstack, Interpreter interpreter,
@@ -71,61 +105,70 @@ class BSHBlock extends SimpleNode
         throws EvalError
     {
         Object syncValue = null;
-        if ( isSynchronized )
-        {
+        if ( isSynchronized ) {
             // First node is the expression on which to sync
             Node exp = jjtGetChild(0);
             syncValue = exp.eval(callstack, interpreter);
         }
 
         Object ret;
-        if ( isSynchronized ) // Do the actual synchronization
-            synchronized( syncValue )
-            {
-                ret = evalBlock(
-                    callstack, interpreter, overrideNamespace, null/*filter*/);
+        if ( isSynchronized )
+            // Do the actual synchronization
+            synchronized( syncValue ) {
+                ret = evalBlock(callstack, interpreter,
+                                overrideNamespace, null/*filter*/);
             }
         else
-                ret = evalBlock(
-                    callstack, interpreter, overrideNamespace, null/*filter*/);
+            // Unsynchronized
+            ret = evalBlock(callstack, interpreter,
+                            overrideNamespace, null/*filter*/);
 
         return ret;
     }
 
-    Object evalBlock(
-        CallStack callstack, Interpreter interpreter,
-        boolean overrideNamespace, NodeFilter nodeFilter )
-        throws EvalError
-    {
+    Object evalBlock(CallStack callstack, Interpreter interpreter,
+                     boolean overrideNamespace, NodeFilter nodeFilter )
+        throws EvalError {
+
         Object ret = Primitive.VOID;
         NameSpace enclosingNameSpace = null;
-        if ( !overrideNamespace )
-            enclosingNameSpace = callstack.swap(
-                    blockspaces.get(callstack.top()));
+        if ( !overrideNamespace ) {
+           /*
+            * Each block gets its own namespace, unless for
+            * some special reason the overrideNamespace is set.
+            * Make a new namespace with the current as a parent
+            * and replace it on to the top of the stack.
+            */
+           enclosingNameSpace= callstack.top();
+           BlockNameSpace bodyNameSpace =
+              new BlockNameSpace( enclosingNameSpace );
+           callstack.swap( bodyNameSpace );
+        }
 
         int startChild = isSynchronized ? 1 : 0;
         int numChildren = jjtGetNumChildren();
 
         try {
             /*
-                Evaluate block in two passes:
-                First do class declarations then do everything else.
-            */
-            if (isFirst || hasClassDeclaration)
-            for(int i=startChild; i<numChildren; i++)
-            {
-                Node node = jjtGetChild(i);
+             * Evaluate block in two passes:
+             * First do class declarations then do everything else.
+             */
+            if (isFirst || hasClassDeclaration) {
+                for(int i=startChild; i<numChildren; i++) {
+                    Node node = jjtGetChild(i);
 
-                if ( nodeFilter != null && !nodeFilter.isVisible( node ) )
-                    continue;
+                    if ( nodeFilter != null && !nodeFilter.isVisible( node ) )
+                        continue;
 
-                if ( node instanceof BSHClassDeclaration ) {
-                    hasClassDeclaration = true;
-                    node.eval( callstack, interpreter );
+                    if ( node instanceof BSHClassDeclaration ) {
+                        hasClassDeclaration = true;
+                        node.eval( callstack, interpreter );
+                    }
                 }
             }
-            for(int i=startChild; i<numChildren; i++)
-            {
+
+            List<Node> enumBlocks = null;
+            for(int i=startChild; i<numChildren; i++) {
                 Node node = jjtGetChild(i);
 
                 if ( node instanceof BSHClassDeclaration )
@@ -138,6 +181,8 @@ class BSHBlock extends SimpleNode
                 // enum blocks need to override enum class members
                 // let the class finish initializing first
                 if (node instanceof BSHEnumConstant) {
+                    if (enumBlocks == null)
+                        enumBlocks = new ArrayList<>();
                     enumBlocks.add(node);
                     continue;
                 }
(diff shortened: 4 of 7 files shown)


nickl-:
> Don't just automagically absorb them from enclosing namespaces, this new way is too surprising and difficult to reason about. It allows third party objects to change how your objects work by setting local variables that will override the closures. If beanshell were a popular scripting language this would be seen as a security hole not a feature.

Not following here, perhaps a code example?

nickl-:
> So this change is [basically a one-liner](https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/main/java/bsh/Name.java#L910), plus changing the test cases to reflect that variable resolution is via the declaring namespace, not the top of the current callstack

We can use Assumes or similar to mark known problem tests to be skipped instead of breaking the test results to fit the solution.

src/main/java/bsh/BshMethod.java
```diff
@@ -345,6 +345,11 @@ private Object invokeImpl(
         else
             localNameSpace = new NameSpace( declaringNameSpace, name );
 
+        // This look dubious because in the case of overrideNameSpace
+        // because will clobber the previous value.  OverrideNameSpace==true
+        // is only used in this codebase by PreparsedScripts and there it is
+        // safe but any other usage might cause problems.  This should be
+        // cleaned up.It was changed as part of #508 but that fix was wrong.
         localNameSpace.isMethod = true;
```
nickl-, line 353: You did not explain why this fix is wrong, was asked to elaborate please.

We are in BshMethod.invokeImpl so these are all methods. Cannot be anything else and never will be. So they should all be marked as isMethod = true. 

How do you fix 508 then?

src/test/resources/test-scripts/scripted_object.bsh
```diff
@@ -0,0 +1,39 @@
...
+   return this;
+}
+
+Object c = null;
+{
+   {
+      c = container();
+   }
+   c.ylds.add(test1("hello"));
+}
```
nickl-, line 25: This I can agree with but if I write 
```
Object c = null;
{
    {
        Object c = container();
    }
    assert(c == null);
}
```

If you redeclare an existing parent variable it must hide the parent. It is how local variable access and inheritance works in JAVA.

Take for example:
```java
jshell> static int getInt() {
...>     int[] i = {1};
...>     class A {
...>         A() {
...>            i[0] = 2;
...>         }
...>     }
...>     new A();
...>     return i[0];
...> }
|  replaced method getInt()

jshell> getInt();
$21 ==> 2
```
Local variable changed in parent namespace.

And then I decide to declare my own variable in the block namespace.
```java
jshell> static int getInt() {
...>     int[] i = {1};
...>     class A {
...>         A() {
...>            int[] i = new int[1];
...>            i[0] = 2;
...>         }
...>     }
...>     new A();
...>     return i[0];
...> }
|  replaced method getInt()

jshell> getInt();
$23 ==> 1
```

Then it does not change the parent which remains 1, but the value in the constructor block is 2 and not 1 that would be wrong, don't you agree?.

nickl-: Inheritance:
```java
jshell> class A { int i = 1;}
|  created class A

jshell> class B extends A {
...>      B() {
...>         i = 2;
...>      }
...>      int getInt() {
...>         return i;
...>      }
...> }
|  replaced class B

jshell> new B().getInt();
$29 ==> 2

jshell> ((A)new B()).i
$30 ==> 2
```
Parent namespace declared the variable so it changed.

But if the child namespace declares the variable:

```java
jshell> class A { int i = 1;}
|  created class A

jshell> class B extends A {
...>      int i = 0;
...>      B() {
...>         i = 2;
...>      }
...>      int getInt() {
...>         return i;
...>      }
...> }
|  replaced class B

jshell> new B().getInt();
$32 ==> 2

jshell> ((A)new B()).i
$33 ==> 1
```

The child namespace value changed but the parent variable was hidden and untouched.

nickl-:
I can completely agree with scripted_object.bsh
```java
test1(String c) {
   return this;
}

container() {
   ylds = new ArrayList();
   return this;
}

Object c = null;
{
   {
      c = container();
   }
   c.ylds.add(test1("hello"));
}
assert("hello".equals(c.ylds.get(0).c));
Object d = null;
{
   {
      d = container();
   }
   d.ylds.add(test1("goodbye"));
}
assert(c.ylds.size() == 1);
assert("hello".equals(c.ylds.get(0).c));
assert(d.ylds.size() == 1);
assert("goodbye".equals(d.ylds.get(0).c));
```

But will this also still be true?

```
test1(String c) {
   return this;
}

container() {
   ylds = new ArrayList();
   return this;
}

Object c = null;
{
   {
      Object c = container();
      c.ylds.add(test1("hello"));
      assert("hello".equals(c.ylds.get(0).c));
   }
   assert(c == null);
}

```

nickl-:
So I am trying to sleep but all I manage to do is dream about namespaces...all over again. =)

I remember why we needed the cache.

Try this (10,000 - 10,000,000):
```java
bsh % _start = System.nanoTime(); for (i=0;i++ < 10000;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
--> $4 = 87744210186405L :long
 3ms
--> void
bsh % _start = System.nanoTime(); for (i=0;i++ < 100000;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
--> $5 = 87762536468325L :long
 14ms
--> void
bsh % _start = System.nanoTime(); for (i=0;i++ < 1000000;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
--> $6 = 87768869490347L :long
 164ms
--> void
bsh % _start = System.nanoTime(); for (i=0;i++ < 10000000;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
--> $7 = 87776434451242L :long
 1421ms
--> void
```
Still nothing to brag about, this is more like it:

```java
jshell> long  _start = System.nanoTime(); for (long i=0;i++ < 10000000L;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
_start ==> 88223537891332
328ms

jshell> long  _start = System.nanoTime(); for (long i=0;i++ < 100000000L;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
_start ==> 88229813804431
328ms

jshell> long  _start = System.nanoTime(); for (long i=0;i++ < 1000000000L;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
_start ==> 88235785781660
540ms

jshell> long  _start = System.nanoTime(); for (long i=0;i++ < 10000000000L;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
_start ==> 88242406316349
2515ms
```

nickl-:
So the cache serves a purpose, and there should be no reason that we can't recreate the namespace (especially if we are doing it everytime) when the cache has expired.

The whole idea behind reference cache was that it should be able to recreate the item when requested and found to have expired. It works for the class cache, I have tested it. I have a suspicion the blocknamespace cache was done in a rush and not implemented properly. There is no reason it shouldn't work and the effort is worth it considering the benefits.

nickl-:
> Try this (10,000 - 10,000,000):

Oh yes and this is with 128GB of RAM, wait till GC starts interfering... but you'll agree.

src/main/java/bsh/BshMethod.java
opeongo: I really should have made this remark as a code review comments and not actually put it in the code.  My bad.

You are correct that always these should be marked as isMethod = true in this place.

My concern was that when you are using the cache that you would get a namespace object and mark it here as a method.  Later on when the object gets recycled it may be used for a non-method block, but the method flag has been set here.  There is no place where the flag gets unset even though the namespace is being recycled.

src/test/resources/test-scripts/scripted_object.bsh
opeongo: > If you redeclare an existing parent variable it must hide the parent.

no disagreement with your first example. 

> Then it does not change the parent which remains 1

No disagreement with the second example

> Parent namespace declared the variable so it changed.

No disagreement with the third example

> The child namespace value changed but the parent variable was hidden and untouched.

No disagreement with the fourth example.

I am not sure what point you are trying to make here.  I am not disagreeing with how Java works.  The issue that I was raising was in regards to the new semantics introduced in to beanshell, and how they differ from java and previous beanshell.

opeongo:
And I should mention another reason that I don't like the new lookup semantics, which I have already mentioned elsewhere several times.  Having variable lookups through the current namespace makes it hard to reason about how the program is going to work, because you have to be aware of every context the the object will be run in.  Accidentally reusing the same variable names somewhere up the the namespace chain can cause subtle and hard to find bugs.  On the other hand, variable lookups through the declaring name space always work the same way, no surprises.

To be kind I guess I could extend my imagination to say that the benefit of looking up variables in the current names space is that you can change behaviour within the scripted just by changing a local variable, so it saves keystrokes.  But I see this as a pretty low value feature.  You can already make these programmatic behaviour changes in a safer and more easily reasoned about way by having setters or using parameters on the scripted object methods.  The kind of accidental mutability that is being exposed here is just a bad idea.  It seems like the mantra these days is "mutability bad", and this change just throws it all open.

So in the balance:
pros
- savings in keystroke perhaps?
- i got nothing else

cons
- scripted objects no longer a closure
   + different from java
   + change from previous beanshell
- more difficult to reason about where values come from
- too much potential for accidental mutation

opeongo:
In the above I have stated that you made these changes, but I did not mean that as an accusation.  It is just that the git history records that you were the one who committed the changes.  If you are not the author of the changes can you let me know what is going on?  Perhaps that will help to make a bit more sense of this.

opeongo:
> > So this change is [basically a one-liner](https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/main/java/bsh/Name.java#L910), plus changing the test cases to reflect that variable resolution is via the declaring namespace, not the top of the current callstack
> 
> We can use Assumes or similar to mark known problem tests to be skipped instead of breaking the test results to fit the solution.

I did not break the the test results to fit my solution.  I posted this PR to help explain the situation as I see it, and for discussion purposes.  I gave my reasoning for why I think the results should be as they are, and I stand by that.

opeongo:
> I remember why we needed the cache.
> 
> Try this (10,000 - 10,000,000):
> ```
> jshell> long  _start = System.nanoTime(); for (long i=0;i++ < 10000000000L;) {} print(((System.nanoTime() - _start) / 1000000)+"ms");
>_start ==> 88242406316349
>2515ms
>```

Ok, so this is about a specific micro benchmark optimization.  And that is a pretty good speed up so worth taking a look at.  But the particular optimization you have chosen is not applicable everywhere.

What your cache is doing is avoiding the alloc/gc costs and recycling an existing namespace.  And in some cases this works (like above), and some cases it fails (like my example).  What is different is that in my case when using scripted objects the `this` reference leaks out and is used elsewhere, so in that case a cached object is not appropriate.  So if you are going to use this caching system then you will need to do an escape analysis to determine when a cached namespace object is safe to use and when it is essential that a fresh namespace object is allocated instead.

Secondly, I think _maybe_ (hedging because I could be proven wrong as always) that organizing the cache as a map using weak references to the caller namespace as the keys is unnecessary and is more trouble than worth.  The namespaces in the cache are all reinitialized to empty so what does it matter to to try to match them up to their callers?  Just keep a collection of recyclabled namespace objects in a synchronized stack.  And the stack contains hard references.  When escape analysis shows that a recycled namespace can be used then pop one off the stack, use it, and push it back when done with it.  Maybe if the stack has gotten too big (say mare than 50 recycled namespace objects?), like after a deeply nested excursions on multiple threads, then just let go of the returned namespace object rather than pushing it.  The stack size upper threshold is something that might be self-tunable (keep track of max stack depth within the past few seconds to see what the pattern is.  All sorts of things to try).  But anyway, using a stack instead of a map would still be sufficient for you to get the performance that you were seeing on your micro benchmark.  And it would avoid all of the trouble and complexity of the ReferenceCache.

Anyway, just some thoughts on how to keep the performance you are getting from namespace recycling and also not break scripted objects etc.

nickl-:
> But regardless of the opinion on how it should be done this is a change from what was working in the previous beanshell version. And it is different from how java does it, and that is surprising. So I think that this change is going to catch people unaware and will perhaps cause a bit of pain for those who were relying on the original behaviour.

I really don't care about how it was working in a beta, only how it should be working and how JAVA works. We have identified #659 there is something wrong with methods returning `this` which needs to me fixed. Already agreed upon so there is no need to keep motivating the fact, it only confuses the issue. 

It is broken and will be fixed, finished and done.

nickl-:
> I did not break the the test results to fit my solution. I posted this PR to help explain the situation as I see it, and for discussion purposes. I gave my reasoning for why I think the results should be as they are, and I stand by that.

So you say that the inherited method **should** return the parent value and not the value at that scope, is this how JAVA works?

https://github.com/beanshell/beanshell/blob/a288337a9ae6ba8329420ca9117fb791f120889f/src/test/java/bsh/Namespace_Chaining_Test.java#L80-L97

We'll get back to this...

nickl-:
> > ```
> > 
> > ```
> 
> 
> Ok, so this is about a specific micro benchmark optimization.  And that is a pretty good speed up so worth taking a look at.  But the particular optimization you have chosen is not applicable everywhere.

I disagree, instead this is quite a generous comparison with an empty blockspace, which we'll hardly ever have in real life, anything more will only compound the situation.

Yes there is work to be done on the cache and we can look at alternative approaches but the performance benefit outweighs the effort needed to resolve this.

nickl-:
> Otherwise something is seriously broken.

I see you did not comment on any of these tests so I take it they are broken now, can we agree?

nickl-:
> We'll get back to this...

This is how JAVA works, which leads me to think the problem is elsewhere, in the way namespace handles fields.

* methods are overwritten
* fields (and variables) are hidden

```java
jshell> class A {
...>     int bar = 42;
...>     int foo() {
...>         return bar;
...>     }
...> }
|  replaced class A

jshell> class B extends A {
...>     int bar = 4711;
...> }
|  replaced class B

jshell> class C extends B {
...>     int bar = 5;
...> }
|  created class C

jshell> (new C()).foo();
$66 ==> 42

jshell> ((B)new C()).foo();
$67 ==> 42

jshell> ((A)new C()).foo();
$68 ==> 42

```

Additional test to pass:

```java

jshell> class B extends A {
...>     int bar = 4711;
...>     int foo() {
...>         return bar;
...>     }
...> }
|  replaced class B

jshell> (new C()).foo();
$70 ==> 4711

jshell> ((B)new C()).foo();
$71 ==> 4711

jshell> ((A)new C()).foo();
$72 ==> 4711

```

opeongo:
> > Otherwise something is seriously broken.
> 
> I see you did not comment on any of these tests so I take it they are broken now, can we agree?

Yes of course this works, it was never in question.  I have updated the test case to include this case.

You could have easily checked this yourself by pulling this branch, it would only have taken a few minutes to see.

opeongo:
> So you say that the inherited method **should** return the parent value and not the value at that scope, is this how JAVA works?

I don't think I ever said that, but if I did I spoke wrongly.   I don't disagree with anything that you are saying about inheritance,  But you are talking about something quite different that what I am talking about.  The issue I am talking about has nothing to do with inheritance.  In fact I don't recall ever mentioning inheritance.  I am talking about closures.  Maybe the some of the confusion was because my previous example wasn't a good one.  I will try again with a another example.

What I have been trying to say is that scripted objects are closures. [You can read about them here](https://beanshell.github.io/manual/objects.html#Scripted_Objects).  When you look up variables in a closure you start in the namespace of the current block (of course) and then if a variable is not resolved you then continue the search in the declaring namespace, just like java does.  

The problem with 3.0-master is that in a scripted object if the variable is not found in the current block then the search continues in the namespace at the top of the callstack, in other words the local variables at the runtime invocation location and so on up through the parent namespace chain.  This is not the same as what happens in Java.

I have reworked my example to show this a bit clearer.  First the java part:
```java
public class ClosureExample {

    public interface Internal {
        void doit();
    }

    public static void main(String[] args) {

        Internal doer;

        {
            int x = 4;
            doer = new Internal() {
                    public void doit() {
                        System.out.println("doing x="+x);
                    }
                };
        }
        doer.doit();
        int x = 5;
        doer.doit();
    }
}
```
and the result of running this is:
```
beanshell>java ClosureExample
doing x=4
doing x=4
```
So obviously the locally declared variable `x` that has a value of 5 is not in the scope of the `Internal` object, so it has no effect.  The `x` variable with a value of 4 is in the declaring namespace where the object was instantiated is the value that is captured by closure contained in `doer`.

And have a look at the equivalent scripted object example.  
```java
doer = null;
 {
     int x = 4;
     internal() {
         doit() {
             System.out.println("doing x="+x);  // Line 6
         }
         return this;
     }
     doer = internal();
 }


doer.doit();  // Line 14
x = 5;
doer.doit();
```
This [BeanShell scripted object](https://beanshell.github.io/manual/objects.html#Scripted_Objects) is mostly identical in function to the Java case.  The `internal` object is instantiated in a declaring namespace that contains a variable named `x` having a value of 4.  

When this is run under beanshell 2.*, or 3.0 plus the patches in this PR the result is like java:
```
beanshell>java -cp target\bsh-3.0.0-SNAPSHOT.jar bsh.Interpreter ClosureExample.bsh
doing x=4
doing x=4
```
This is because the namespace for this closure starts with the local variables, and then continues to the declaring namespace.  The variable named `x` was not found in the namespace for the `internal` object, so the search continues in the **declaring namespace**, and the value of 4 is found.

When this script is run using the code currently in 3.0 master the result is:
```
Evaluation Error: bsh.EvalError: Sourced file: beanshell\ClosureExample.bsh : illegal use of undefined variable, class, or 'void' literal : at Line: 6 : in file: beanshell\ClosureExample.bsh : ) ;

Called from method: doit : at Line: 14 : in file: beanshell\ClosureExample.bsh : doer .doit ( )
```
This happens because the doit method fails to find a variable named `x` starting in the local variables to the method.  So it continues to **the namespace at the top of the callstack** which is the place where `doit` was called from.  There is no variable named `x` in that context yet, so there is an error.

If the statement at Line 14 is commented out then at Line 15 a variable named `x` is set to a value of 5 in the local namespace.  Rerunning the script yields the following result:
```
doing x=5
```
As you can see, this value is coming from the local variables at the invocation site.  This is not what java does.

I don't know how else to explain this issue.  Other to say again that it is not about inheritance.  It is about the namespace search chain for resolving variables in closures.

nickl-:
Still working on this as time allows...

src/main/java/bsh/BshMethod.java
nickl-: You were right, it was making a real mess there, not even sure why it resolved the issue either. Even the global namespace was marked as a method. Well spotted on the cached namespaces, it is not enough for unique name spaces to only use the parent namespace. Subsequent blocks are bungled up and yes inherit the method flag too. This certainly was not the intention...

I have tucked this flag set back in under the else where it was and added a unique block id to the block name space cache, per block instance which now appears to behave like it should. Creates a new namespace per unique block and only reuses it for that specific block, if needed for recursion, loops, etc. The performance still seems on par with the original implementation and the namespaces are not being polluted anymore.

I tried to merged all your other fixes in from this branch, will have to see if I missed anything.

nickl-:
> I don't know how else to explain this issue. Other to say again that it is not about inheritance. It is about the namespace search chain for resolving variables in closures

I don't think it has anything to do with lexical scope, unless it also means something else. The lexicon or code defines the call stack which is visible from where bindings can be found. I have difficulty trying to understand you as lexical scope remains the same, with or without late binding (or whatever we are calling it)

Perhaps what you are trying to distinguish, and I am grasping at straws so forgive me if I am wrong, is the difference between static and dynamic scope. Java is compiled (as with all compiled languages) and requires that everything is known at the point of declaration, once compiled it is hard coded and in place already attached to the bindings it has discovered, long before execution.

Scripting languages are dynamic they only get parsed at the point of declaration for syntactic correctness, no attempt is made to find bindings and in fact you can declare those afterwards and the block will not be any wiser. In beanshell we strip the page of methods and their code blocks ends up in a map where we can look them up by name.

Only when evaluated do we start looking for bindings, and we use the lexical scope through the call stack and namespace hierarchy (which are lexically defined and populated) to achieve this. We cannot force the scope statically anymore, this will break lexically because the declaration scope might not be visible anymore, nor should we want to because this is a benefit of being a dynamic language only interpreted when executed.

I know you are trying to break it to prove that there is a problem, but also consider this logically. Why would you code something like this for example:

```java
doer.doit(); 
x = 5;
doer.doit();
```

If you want `x = 5` to be ignored? It makes no sense, or am I missing something? Instead I would argue that we should make every effort to incorporate that into the dynamic execution, and if you didn't want `x = 5` because you are relying on another reference in the lexical scope then you won't be polluting the scope and risk it being obfuscated, by giving `x = 5` a more descriptive and unique name to explain its intent and purpose.

So to summarize: JAVA doesn't use closures but classes, and classes gives us a lexical boundary and practical grouping so we can make every effort to dynamically execute the code as JAVA does statically. However JAVA doesn't have closures so it cannot provide an appropriately defined stipulation for us to follow. In this regard we have long left the station or we will soon be pressed to justify the type system to name one, which has a more defined conformity to abide by in JAVA, but we won't. Because it is about the best of both worlds, productivity and simplicity, being allowed to bend the rules when you want to and enabled when you simply have to because there is no other choice.

I am not unreasonable and if something breaks that shouldn't then we will fix it, as we do all the time. But if a developer wants to only use `c` for all their variable names then there is going to be limit to how much we care.

PR #673, opened 2022-12-16
ASM version 9.4 upgrade.

### LINKED ISSUE #656: Upgrade ASM Api
We are currently on v6.1.1 as per #68 

According to the [project page](https://asm.ow2.io/versions.html) the latest version is: v9.4

We include the ASM source in BeanShell instead of a maven dependency.

The following actions need to be considered:
- Applied fixes as per changes to ASM api including deprecated method changes to visitMethodInsn
- As before we still don't need ClassReader so gutted it out of the ASM API.
- We also don't need Annotation or Module support so that got purged too.

nickl-:
Applied the upgrade to 9.4, gutted all unused APIs. Added to new branch for future use. Fixes #656

### DIFF
--- a/src/main/java/bsh/org/objectweb/asm/Attribute.java
+++ b/src/main/java/bsh/org/objectweb/asm/Attribute.java
@@ -28,12 +28,12 @@
 package bsh.org.objectweb.asm;
 
 /**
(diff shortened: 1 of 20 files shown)


nickl-:
Close for now, perhaps for the future... see #656

PR #678, opened 2022-12-23
Fix cloneable support.

### LINKED ISSUE #421: clone is broken
clone does not work, as can be seen in this example:

```java
// clone.bsh
public class A implements Cloneable {
    String name;
    void whoami() {
        System.out.println("I am " + this + " my name is " + name);
    }
}

A a = new A();
a.name = "a";
a.whoami();

A b = a.clone();
b.name = "b";
b.whoami();
// end of clone.bsh
```
The output is:
```java
% java -cp bsh.jar bsh.Interpreter clone.bsh
I am A@589e56 my name is a
I am A@589e56 my name is a
```
This is with 2.0b4 and JVM 1.5.

\- Godmar

Reported by: gback

Original Ticket: "beanshell/bugs/283":https://sourceforge.net/p/beanshell/bugs/283

nickl-:
Fixes #421 which originally reported the problems with cloning scripted objects. 
Had to deep copy all the contexts and most references to finally be able to produce two independent instances after a clone. Also completed the existing attempt to implement cloning, and we now return an actual instance instead of just copied This namespace.

### DIFF
--- a/src/conf/version.properties
+++ b/src/conf/version.properties
@@ -14,6 +14,6 @@
 #
 #Project version number - populated by maven
 #Project build number - incremented by beanshell-maven-plugin
-#Wed Dec 21 23:25:32 SAST 2022
-build=1677
+#Fri Dec 23 02:45:54 SAST 2022
+build=1852
 release=${project.version}
--- a/src/main/java/bsh/BshMethod.java
+++ b/src/main/java/bsh/BshMethod.java
@@ -24,7 +24,6 @@
  *                                                                           *
  *****************************************************************************/
 
-
 package bsh;
 
 import java.util.stream.IntStream;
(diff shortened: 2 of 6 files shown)


nickl-:
Scrutinizer insists that we should [remove the clone](https://scrutinizer-ci.com/g/beanshell/beanshell/inspections/1ea6cb42-793f-4dba-ba67-d0d1df70bc9f/issues/files/src/main/java/bsh/BshMethod.java?status=new&orderField=path&order=asc&honorSelectedPaths=0) I added on bshmethod. There is some poetic justice in the thought that when implementing clone you will at least add one cloneable class.

This patch already took way more time then I initially expected, the methods need to copied or they simply produce garbage, and yes clone is a messed up api and we are likely going to see more issues on this in the future. I will probably write a manual copy for methods at some point but right now it seems like too much effort and no reward. The clone is working for now...

PR #681, opened 2022-12-26
Add reset interpreter method

### LINKED ISSUE #413: Memory leak profiler inspection
Running bsh-204b version whith jdk 1.4, I'm getting a 
memory leak when calling bsh.eval\(\)
Whith a profiler I can see that after the execution o 
bsh, there are still around 25 Mb in memory, most of 
them in unclaimed bsh.Token objects.
As said in the manula: "...BeanShell Interpreter 
instances were designed to be very light weight. 
Construction time is usually negligible and in simple 
tests, we have found that it is possible to maintain 
hundreds \(or even thousands\) of instances."...
Is there a way to free those Tokens?
The only solution is invoke the Garbage Colector?
Thanks

Reported by: convenio

Original Ticket: "beanshell/bugs/275":https://sourceforge.net/p/beanshell/bugs/275

nickl-:
Interpreter allows for multiple `eval` calls and for this purpose certain state is kept. Added reset method to release additional resources and when completely done with interpreter. fixes #413

### DIFF
--- a/src/conf/version.properties
+++ b/src/conf/version.properties
@@ -14,6 +14,6 @@
 #
 #Project version number - populated by maven
 #Project build number - incremented by beanshell-maven-plugin
-#Mon Dec 26 11:05:42 SAST 2022
-build=1915
+#Mon Dec 26 23:27:04 SAST 2022
+build=2015
 release=${project.version}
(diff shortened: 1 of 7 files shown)


nickl-:
Fixed [scrutinizer raised issue](https://scrutinizer-ci.com/g/beanshell/beanshell/inspections/55beee75-c63c-4229-b4de-ed15054734ed) and repush

Issue #727, opened 2023-02-23
Variable scoping compatibility mode


opeongo:
BeanShell 3.0.0 has made a change to the rules for scoping loosely defined variables.  In BeanShell version 2 loosely defined variables were defined in the enclosing scope.  According to the BeanShell documentation:

> Blocks are statements between curly braces {}. In BeanShell, as in Java, blocks define a level of scope for typed variables: typed variables declared within a block are local to the block. Other assignments within the block occur, as always, wherever the variable was defined.
> Untyped variables in BeanShell, however, are not constrained by blocks. Instead they act as if they were declared at the outer (enclosing) scope's level. With this in mind, BeanShell code looks just like Java code. In BeanShell if you declare a typed variable within a block it is local to the block. But if you use an untyped variable (which looks just like an ordinary assignment in Java) it behaves as an assignment to the enclosing scope. 

For example while using BeanShell 2 the following code the assertion will pass because the `x` variable is declared in the outer namespace.
```java
{
  x = 5;
}
assert(x==5);
```

In BeanShell 3.0.0 untyped variables are always declared within the block where they are encountered, behaving more like Java.  Using the same code in BeanShell 3 the assertion will fail because the `x` variable has been declared within the block, and that scope has gone out of context.  In the assert statement `x` will be undefined.

Regardless of how you feel about the sensibility of the earlier behaviour, the reality is that there exists legacy BeanShell V2 code that relies upon the previous behaviour for this feature.  When running old V2 code using a version 3 interpreter errors will occur.  Although the best long-term fix is to edit the code to adapt to the new behaviour, some users might not have the programming capability to make these updates.  For this segment of the BeanShell user base a compatibility mode would help with the transition.  

A Bsh2 variable scoping compatibility mode could be used restore the previous behaviour and allow old scripts to run unchanged.  The compatibility mode could be activated by calling a method on the interpreter (such as `interpreter.setBsh2ScopingCompatibility(true);`), or by setting a command line flag (such as `-Dbsh.Bsh2ScopingCompatibility=true`),

nickl-:
We have had this discussion already: https://github.com/beanshell/beanshell/issues/555#issuecomment-559629112

And the change will be documented as per #555 

Also there is a simple work around:

```java
{
  super.x = 5;
}
assert(x==5);
```

Closed: Won't break again it is fixed

nickl-:
From #720

>  The point is that there is a large code base that I can not access to make changes.

[...]

PR #728, opened 2023-02-23
Implementation for BeanShell 2 compatibility mode (issue #727)

opeongo:
Implement a mode to allow backward compatibility with BeanShell 2 scripts that use scoping rules that declare loosely typed variables in the outer (enclosing) scope (#727).

Although this touches a lot of files, it is actually not very invasive.  The active change is a few lines in [NameSpace.java](https://github.com/beanshell/beanshell/blob/5ae59f477f9f9a889ca766aaf4083ba97e67f7ee/src/main/java/bsh/NameSpace.java#L426).   The bulk of the changes are to change the implementation of the strictJava flag from a boolean to an int flag container such as argument passing and value testing.


nickl-:
Closed as per #727

PR #729, opened 2023-02-27
Fix for issue 725 casts on numeric references types causing exceptions

opeongo:


nickl-:
Looks good but we are loosing too much code coverage. Please add tests for the other paths too:

Some or all of these conditions are missing:

1.
```java
fromType == null
```
2.
```java
fromType.isPrimitive()
```
3.
```java
!toType.isAssignableFrom( fromType )
```
4.
```java
checkOnly == true
```

I know it is getting harder to maintain or even improve on the overall coverage, but the unit tests only cover a 3rd of this patch.

opeongo:
> Looks good but we are loosing too much code coverage. Please add tests for the other paths too:
> 
> Some or all of these conditions are missing:
> 
> 
> ```java
> fromType == null
> ```
> 
> 
> ```java
> fromType.isPrimitive()
> ```
> 
> 
> ```java
> !toType.isAssignableFrom( fromType )
> ```
> 
> 
> ```java
> checkOnly == true
> ```
> 
> I know it is getting harder to maintain or even improve on the overall coverage, but the unit tests only cover a 3rd of this patch.

It is not easy to write tests for all of these conditions.  For example, how to I write a test for checkOnly when it is not exposed to the scripting environment?

nickl-:
> It is not easy to write tests for all of these conditions. For example, how to I write a test for checkOnly when it is not exposed to the scripting environment?

Surely we can write a test with the appropriate conditions to cover the condition without it being specifically exposed to the scripting environment. If we can't cover it then why is it implemented then, perhaps then we don't need the condition?

opeongo:
> > It is not easy to write tests for all of these conditions. For example, how to I write a test for checkOnly when it is not exposed to the scripting environment?
> 
> Surely we can write a test with the appropriate conditions to cover the condition without it being specifically exposed to the scripting environment. If we can't cover it then why is it implemented then, perhaps then we don't need the condition?

My guess is that the code is there as a sanity check to ensure that the contract is being met.  It would take a lot of auditing to make sure that those tests can be safely removed.

I have no more tests to add for this one that will cover those existing code paths.

Issue #731, opened 2023-03-13
Method lookup regression


opeongo:
Here is an example of a regression of a method lookup that previously worked with BeanShell 2.0b5 and no longer works in BeanShell 3.0.  

This following code works in BeanShell-2.0b5, but fails in BeanShell-3.0.0:
```java
method1( single ) {
   print("in method1(single)");
}

method1( String[] list ) {
   for( int i=0; i<list.length; i++)
         method1( list[i] );
}

method1("a");
```

When run under BeanShell-3.0.0 the following error is raised:
```
beanshell>java -cp target\bsh-3.0.0-SNAPSHOT.jar  bsh.Interpreter poly.bsh
Evaluation Error: bsh.EvalError: Sourced file: xxxx\poly.bsh : illegal use of undefined variable, class, or 'void' literal : at Line: 6 : in file: xxxx\poly.bsh : ;

Called from method: method1/BlockNameSpace3 : at Line: 10 : in file: xxxx\poly.bsh : method1 ( "a" )
```

There is a simple workaround to this issue which is to change the first method declaration to `method1(String single) {`.

Also, if the single string version of `method1` is declared after the array version this will also work.  This works correctly:
```java
method1( String[] list ) {
   for( int i=0; i<list.length; i++)
         method1( list[i] );
}

method1( single ) {
   print("in method1(single)");
}

method1("a");
method1(new String[] {"a", "b", "c"});
```

I don't consider this to be a significant problem for me because I always use formal parameters with declared types.  I am bringing this up as an issue because it may relate to other lookup problems that have been showing up lately.  Also, it is odd to me that the error is occurring at line 6 which should not be on the execution path.  It appears as though the wrong code block is being executed.

opeongo:
I traced through the code to see where the problem is and the mixup is happening in [Types.castObject line 559](https://github.com/beanshell/beanshell/blob/47cbe833b7709fe0c5b6c67ee51b731746a7bfe5/src/main/java/bsh/Types.java#L559).  This problem was introduced to the code base 5 years ago, so we have a testing problem.

[...]

PR #732, opened 2023-03-14
Changes related to issue #731 method lookup regression. 

opeongo:
This is only a partial fix because each change uncovered another subtle issue.  Still work to be done.   I am posting this PR so that someone else who has a better idea of how this is supposed to work can at least see some of the issues that I found.

Fixing the issue for proper matching of the most specific method broke several other code paths that depended on the loose behaviour of the `Types.castObject` method.
Other issues that popped up that I also fixed included:
- array dimensions occurring after a variable name being ignored (hidden problem that had not been tested yet).
- arrays of different dimensions but same base type being incorrectly matched as being cast compatible.
- Arrays of collections being cast to collections (surfaced by the change to `Types.castObject`, wasn't a problem before)

Unfortunately as I made changes more issues started to crop up and I just ran out of time.  For example
- `new {{false}, true};` no longer compiles and executes correctly.  This looks to me like it shouldn't work anyway, because it is not a proper 2D array. 
- `entries = {{new Entry {"k"=3, "l"=5}, {6}}};` should get typed as `Object[][][]` but now is getting typed as 'MapEntry[][][]` and this fails.
I suspect that I might have been going down the wrong track with some of my changes.  Perhaps there is a simpler way to fix the original problem that does not cause so much other collateral damage.

### DIFF
--- a/src/main/java/bsh/BSHArrayInitializer.java
+++ b/src/main/java/bsh/BSHArrayInitializer.java
@@ -111,7 +111,7 @@ class BSHArrayInitializer extends SimpleNode {
 
         // force MapEntry to Map output
         if (dimensions < 2)
-            if ( MapEntry.class == inferType && Void.TYPE == baseType
+            if ( (MapEntry.class == inferType && Void.TYPE == baseType)
                  || MapEntry.class == baseType )
                 baseType = Map.class;
 
--- a/src/main/java/bsh/BSHArrayInitializer.java
+++ b/src/main/java/bsh/BSHArrayInitializer.java
@@ -26,6 +26,7 @@
 package bsh;
 
 import java.util.ArrayDeque;
+import java.util.Collection;
 import java.util.Deque;
 import java.util.Map;
 import java.lang.reflect.Array;
@@ -251,10 +252,15 @@ class BSHArrayInitializer extends SimpleNode {
      * @throws EvalError thrown on cast exceptions */
     private Object toCollection(Object value, Class<?> type, CallStack callstack)
             throws EvalError {
+        /*
+         * Don't try to convert an array of collections into a collection
+         */
         Class valClaz = value.getClass();
+        Class valBase = Types.arrayElementType(valClaz);
         if ( Types.isCollectionType(type) &&
              !(valClaz.isArray() &&
-               type.isAssignableFrom(Types.arrayElementType(valClaz))))
+               (Map.class.isAssignableFrom(valBase) ||
+                Collection.class.isAssignableFrom(valBase))))
             try {
             return Types.castObject(value, type, Types.CAST);
         } catch ( UtilEvalError e ) {
--- a/src/main/java/bsh/BSHArrayInitializer.java
+++ b/src/main/java/bsh/BSHArrayInitializer.java
@@ -109,9 +109,10 @@ class BSHArrayInitializer extends SimpleNode {
             inferType = inferCommonType(null, this, callstack, interpreter);
 
         // force MapEntry to Map output
-        if ( MapEntry.class == inferType && Void.TYPE == baseType
-                || MapEntry.class == baseType )
-            baseType = Map.class;
+        if (dimensions < 2)
+            if ( MapEntry.class == inferType && Void.TYPE == baseType
+                 || MapEntry.class == baseType )
+                baseType = Map.class;
 
         // no common type was inferred
         if ( null == inferType ) {
@@ -145,6 +146,7 @@ class BSHArrayInitializer extends SimpleNode {
         // the values are set.
         dims[0] = jjtGetNumChildren();
         Object array =  Array.newInstance( baseType, dims );
+        Class<?> entryType = array.getClass().getComponentType();
 
         // Evaluate the child nodes
         for ( int i = 0; i < jjtGetNumChildren(); i++ ) {
@@ -180,7 +182,7 @@ class BSHArrayInitializer extends SimpleNode {
             try {
                 // store the value in the array
                 Array.set(array, i,
-                        normalizeEntry(entry, baseType, dimensions, callstack));
+                        normalizeEntry(entry, entryType, dimensions, callstack));
             } catch( IllegalArgumentException e ) {
                 Interpreter.debug("illegal arg", e);
                 throwTypeError( baseType, entry, i, callstack );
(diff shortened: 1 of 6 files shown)


opeongo:
I took another look and was able to fix the remaining issues, the fix just had to gestate.  By the way `new {{false}, true};` was explicitly test and checking the for an error, and my fix changed the error message (I think it made it better).  Changing the expected error message fixed that part.

This PR should be ready to merge.

src/main/java/bsh/BSHFormalParameter.java
```diff
@@ -62,9 +62,24 @@ public String getTypeDescriptor(
...
+             * - If there are already array dimensions on the type then
+             * throw an error.
+             */
+            if (dimensions > 0) {
+                if (!type.isArray())
+                    type = Array.newInstance(type, new int[dimensions]).getClass();
+                else
+                    throw new EvalError(
+                       "Array dimensions not allowed on both type and name: "
+                       + name, this, null );
```
nickl-, line 80: This seems like an easy enough test to add.

src/main/java/bsh/BSHArrayInitializer.java
```diff
@@ -109,9 +110,10 @@ public Object eval( Class<?> baseType, int dimensions, CallStack callstack,
             inferType = inferCommonType(null, this, callstack, interpreter);
 
         // force MapEntry to Map output
-        if ( MapEntry.class == inferType && Void.TYPE == baseType
-                || MapEntry.class == baseType )
-            baseType = Map.class;
+        if (dimensions < 2)
+            if ( MapEntry.class == inferType && Void.TYPE == baseType
+                 || MapEntry.class == baseType )
```
nickl-, line 115: Seems like a fair suggestion, I know the code didn't change but it came under review now.

src/main/java/bsh/BSHArrayInitializer.java
```diff
@@ -109,9 +110,10 @@ public Object eval( Class<?> baseType, int dimensions, CallStack callstack,
             inferType = inferCommonType(null, this, callstack, interpreter);
 
         // force MapEntry to Map output
-        if ( MapEntry.class == inferType && Void.TYPE == baseType
-                || MapEntry.class == baseType )
-            baseType = Map.class;
+        if (dimensions < 2)
```
nickl-, line 113: Why does the dimensions matter again?

src/main/java/bsh/BSHArrayInitializer.java
```diff
@@ -145,6 +147,7 @@ private Object buildArray(int dimensions, Class<?> baseType,
         // the values are set.
         dims[0] = jjtGetNumChildren();
         Object array =  Array.newInstance( baseType, dims );
+        Class<?> entryType = array.getClass().getComponentType();
```
nickl-, line 150: Isn't this what `arrayElementType` is supposed to do? I am still curious if that is not broken.

src/main/java/bsh/Types.java
```diff
@@ -556,7 +556,7 @@ If fromValue is (or would be) Primitive.NULL then fromType should be null.
     public static Object castObject( Class<?> toType, Class<?> fromType, Object fromValue,
             int operation, boolean checkOnly ) throws UtilEvalError {
         // assignment to loose type, void type, or exactly same type
-        if ( toType == null || arrayElementType(toType) == arrayElementType(fromType) )
```
nickl-, line 559: What happens if we fix `arrayElementType`, I am wondering if this does not cause problems elsewhere...

src/main/java/bsh/Types.java
```diff
@@ -556,7 +556,7 @@ If fromValue is (or would be) Primitive.NULL then fromType should be null.
...
+        if ( toType == null || toType == fromType )
```
nickl-, line 559: However this condition does satisfy the comments, for what that's worth.

src/test/resources/test-scripts/expression_array.bsh
```diff
@@ -340,7 +340,7 @@ assert(isEvalError("Error array get index: Index 3 out-of-bounds for length 2",
 assert(isEvalError("Not an array or List type", "''[1];"));
 assert(isEvalError("Class: Unknow.ObjectType not found in namespace", "new Unknow.ObjectType[0];"));
 assert(isEvalError("Incompatible initializer. Allocation calls for a 2 dimensional array, but initializer is a 1 dimensional array", "new int[][] {};"));
-assert(isEvalError("Incompatible type: boolean in initializer of array type: boolean", "new {{false}, true};"));
+assert(isEvalError("Error in array initializer: Cannot cast primitive value to object type [Z", "new {{false}, true};"));
```
nickl-, line 343: Agreed more precise error now.

nickl- (changes requested):
More nitpicking than much else, over all looks good. Just worried about `arrayElementType`...

src/main/java/bsh/BSHArrayInitializer.java
nickl-: Reference to Lift's suggestion, now seems to be lost.

> [OperatorPrecedence](https://errorprone.info/bugpattern/OperatorPrecedence): Use grouping parenthesis to make the operator precedence explicit

src/main/java/bsh/BSHFormalParameter.java
nickl-: Codecov flagged as untested.

src/main/java/bsh/BSHArrayInitializer.java
opeongo: It cannot be coerced to a map if there are more than two dimensions

src/main/java/bsh/BSHArrayInitializer.java
opeongo: Ok, added

src/main/java/bsh/BSHArrayInitializer.java
opeongo: Do you mean Types.arrayELementType()?  That will return the ultimate base type of the array, so for int[][] it will return int.  However here we want the type for the current dimension, which would be int[].

src/main/java/bsh/BSHFormalParameter.java
opeongo: Added test

src/main/java/bsh/Types.java
opeongo: You cannot just test for equality by comparing the ultimate base types because the number of dimensions might be different.

src/main/java/bsh/Types.java
opeongo: Comment changed

opeongo (commented):
Changes made

opeongo:
Hopefully this can be merged

nickl-:
> Hopefully this can be merged

Been a while, will need to pick up again to see where we are at.

opeongo:
Has this been merged?  If so should be closed.

PR #741, opened 2023-06-01
Fix for #659 variable lookup confusion

opeongo:
- Change the way the way that the block namespace cache is implemented.  Previously it used an imprecise equals method that led to collisions and occasionally returned null values.

- Move the namespace swap out of the iterative loop in BSHEnhancedForLoop to avoid unnecessary lookups.


opeongo:
Can we have some discussion on this or perhaps get it merged?

jrabun:
Just one anecdotal data point:  My application was encountering the same NPE mentioned in [this comment](https://github.com/beanshell/beanshell/issues/659#issuecomment-1567515980).  Applying the changes from this branch resolved the issue.  Thank you, opeongo, for the time and effort you put into debugging this issue!

opeongo:
> Just one anecdotal data point: My application was encountering the same NPE mentioned in [this comment](https://github.com/beanshell/beanshell/issues/659#issuecomment-1567515980). Applying the changes from this branch resolved the issue. Thank you, opeongo, for the time and effort you put into debugging this issue!

Thanks for the feedback.  It is nice to know that someone else is out there.

nickl-:
Will take some time to review this, it seems to be doing more than one thing.

Issue #750, opened 2023-12-05
Request to fix broken original author info...


patniemeyer:
Hi everyone - I’m Pat Niemeyer, the original author of BeanShell from way back.  I apologize in advance for asking this since it’s an especially annoying request, but I have figured out why Github doesn’t associate me with the project as an author/contributor.   When @nickl-   did the initial import of the code from the old SVN repository, the author email got mangled with some kind of uid suffix, so GitHub can’t find the author to link.

Unfortunately the only way to fix this is to do some surgery on that commit with `git filter branch` and force push the change.  I even pinged GitHub support to see if they had any ideas on a less invasive way but that is all they could come up with.

I recognize that doing this would mess with working copies that developers have cloned locally and would be a major inconvenience for them, but if there is a time in the future when we could fix this I would greatly appreciate it.  I am active on Github and it would be nice to be associated with this project that I spent so much time on in years past :) 

For reference, this is the raw commit info that shows the broken email:
https://github.com/beanshell/beanshell/commit/c4fc63a1e39e0637ac17090e4f27c0aec95cc2b9.patch

And this is a suggested script that should fix it.
```
# Clone a bare copy of the repo
git clone --bare https://github.com/<username>/<reponame>.git
cd <reponame>.git

# Fix up the broken email
#!/bin/sh
git filter-branch --env-filter '
BAD_EMAIL="patniemeyer@gmail.com@934af587-6f8e-29cc-0aa7-85b2284b99e2"
CORRECT_EMAIL="patniemeyer@gmail.com"
if [ "$GIT_AUTHOR_EMAIL" = "$BAD_EMAIL" ]; then
    export GIT_AUTHOR_EMAIL="$CORRECT_EMAIL"
fi
' --tag-name-filter cat -- --branches --tags

# Double check that it went well and then:
# git push --force --tags origin 'refs/heads/*'
```

nickl-:
Please not there are some new PRs merged since my previous comment so pull fresh HEAD before making changes.

opeongo:
> I recognize that doing this would mess with working copies that developers have cloned locally and would be a major inconvenience for them

Could you provide a bit more information about what this means?

I have a local copy of the repository and I just did a pull from origin/master and many branches got forced updates.  Are there some steps that I need to take to resolve this before I mess things up further?  What will happen to my outstanding pull requests?

opeongo:
[...]

Issue #645, opened 2022-06-11
Restrict access to specific classes and any keyword that we mention(custom defined keyword)


KameshwarB:
Hello !!!
in my project i want to restrict certain imports and keywords not to be use by script, therefore when it sees it terminates and throw exception.
Let's say we have class imports from **java.sql. or spring.jpa. or spring.jdbc. or any code defined keywords** we want to restrict beanshell from executing the script and throw exception.

is it possibile?

nickl-:
Interesting. How would we implement this with standard java security policy? ...it should work as expected.

nickl-:
Also see #35

nickl-:
Closed - lack of further discussion assumes security policy resolved the issue

PR #751, opened 2023-12-12
SecurityGuard: An API to prevent malicious code execution

Net-0:
Hello guys,
 
  I work as a software developer in Brazil ( I apologize for my english xD ), and the product that I work with my team uses the BeanShell to execute simple Java Snippets, but we were doing security validations of the entire system, and we were minding about how to solve the security problem of malicious code executions, e.g. executing "System.exit(1)", it'd be a really problem to us. Then I read a lot about "eval" functionalities, e.g. the JavaScript **eval** function, and I understand that it's a concern that "evals" are unsecure when you don't know what are being executed, so I propose of creating an API that solve this problem ( at least to Java :P ), and I called it **SecurityGuard**.
 
  "Ok, ok, but how does it work?", the idea is really simple, the **SecurityGuard** is just an interface with methods that validate if some specific thing can be executed, e.g. the interface has a method called **canInvokeStaticMethod** that you can implement to validate the variables to call a method and return a boolean saying if it can be executed or not. E.g. of a code implementation of **SecurityGuard** that prevent the execution of "System.exit(1)":
image

  "And after that?", then you just need execute "Interpreter.mainSecurityGuard.add(mySecurityGuard)", to add your guard to prevent the execution of some dangerous code **for your application**.
  
  "What occurs when my SecurityGuard says that something can't be executed?", then the Interpreter just throws a **SecurityError** ( it's a new error that extends **UtilEvalError**, and is used at the same way in the code )

  "And it impacts performance?", my team did this question when I present the POC of this idea, and the answer is quite simple: it dependes of how much SecurityGuards the Interpreter has and the validation of each of them, but with a small amount of SecurityGuards that aren't over complex, it'll be a small performance change. E.g. with 50 simple SecurityGuards, the performance impact is about 2%, I can redo the tests and send you the results if you wanna :).
  
  **Disclaimers**
- Exists an implementation of **SecurityGuard** called **BasicSecurityGuard**, that just prevent that the code being executed by BeanShell add/remove SecurityGuards from the **Interpreter.mainSecurityGuard**, because it'd create another security problems xD.
- Reflection results at normal validations, e.g. if the code call some method using Reflection, the method **canInvokeMethod** is called yet!!
- I'd rather that the 'mainSecurityGuard' be for each instance of Interpreter, but in the v3 it's really hard to implement it ( at least to me :V ) because of how BeanShell works, and it'll also create a problem that the code being executed by BeanShell is able to create an instance of Interpreter to execute the malicious code, and there is an example saying that is possible to create an instance of Interpreter in BeanShell page, so it's a hard way to do it.
- I'm a Jr. Software Developer yet, and it's my first contribution at open source project, so any feedback is welcome ;D
- My team didn't use SecurityGuard, this solution is powerful and amazing to our case, but my tech leader said that we mustn't maintain a local fork of an entire project just because we need some small changes in the source code, so if this fork be approved, we should use it, otherwise no :P, meanwhile we're using another validation that isn't as precise but is better than nothing.


nickl-:
Sounds interesting.
 
You will need to roll this against fresh HEAD, we had to do some maintenance on the repo at #750, I believe your changes were done on a previous copy which is why it shows so many changes.

Please redo the PR with only the changes for the `SecurityGuard` for a proper revue.

nickl-:
May relate to  #645

Net-0:
Hello, sorry for being late xD, I just saw your comment yesterday.

I rebased my fork's branch with the original master, and it seems ok now ;D

Net-0:
Hello @nickl- , does the pull request have some issue 😬 ? It wasn't approved yet...

Net-0:
Hello nick,

No, it doesn't have a mean to prevent commands, to be honest, I guessed that it didn't need it xD, but reading the all available commands, I think that it can be interesting for some specific cases, I'm going to do a implementation for this purpose :).

Net-0:
Hello bro,

Now the SecurityGuard have a mean to prevent the execution of local methods/commands :)

Good Night ✌️

nickl-:
Awesome!!

We have a public holiday on Thursday, will have time then to give my full attention.

src/main/java/bsh/security/SecurityGuard.java
```diff
@@ -0,0 +1,37 @@
+package bsh.security;
+
+/** It's the interface to implement a single SecurityGuard to be used by MainSecurityGuard */
+public interface SecurityGuard {
```
nickl-, line 4: Not sure if this needs to be public either.

nickl-:
Please also see Scrutinizer exceptions to rectify.

src/main/java/bsh/security/MainSecurityGuard.java
```diff
@@ -0,0 +1,224 @@
...
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import bsh.Interpreter;
+
+/** It prevents the execution of codes that manipulate the SecurityGuard or MainSecurityGuard */
+class BasicSecurityGuard implements SecurityGuard {
```
nickl-, line 14: Could these not be default methods on the interface, to be overwritten by custom security guard?

nickl-: In other words, if the methods of `BasicSecurityGuard` are implemented as default methods on interface `SecurityGuard` then we can remove the need for the `BasicSecurityGuard` class.

nickl-:
Implementation otherwise looks good and this will be a very useful addition. 
I should have time again Monday or Tuesday if this also works for you.

src/main/java/bsh/Types.java
```diff
@@ -43,7 +43,7 @@
 
 
 */
-class Types {
+public class Types {
```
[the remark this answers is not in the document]
Net-0: pretty fair, I didn't think about this class accessor, I'm moving out the Security classes of the package `bsh.security`.

src/main/java/bsh/security/SecurityError.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.List;
+
+import bsh.CallStack;
+import bsh.EvalError;
+import bsh.SimpleNode;
+import bsh.Types;
+import bsh.UtilEvalError;
+
+/** It's a specific error that is throwed when try to execute something that mustn't be executed */
+public class SecurityError extends UtilEvalError {
```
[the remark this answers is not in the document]
Net-0: hm... ok, it seems fine to me, but for curiosity, can I ask why UtilEvalError is public ? Just to ensure backwards compatibilities ?

src/main/java/bsh/security/SecurityGuard.java
Net-0: Yep, it really needs to be public, this interface must be implemented by users if them wanna prevent the execution of some specific thing, so them need to implement this interface just with their validation logic, then just add their implementation with `Interpreter.mainSecurityGuard.add()`

src/main/java/bsh/security/MainSecurityGuard.java
Net-0: I guess no, the `SecurityGuard` is just a generic interface, the validation logic to block the execution of something still must be implemented, the `BasicSecurityGuard` is an internal logic validation to prevent that code interpreted by BeanShell can interact with the API of Security ( add or remove a `SecurityGuard` ).

But I can put a default implementation to `SecurityGuard` methods to avoid this boilerplate of `return true;` at so many methods.

Net-0:
> Please also see Scrutinizer exceptions to rectify.

Abou this, I fixed the code style issues of Scrutinizer.

The tests results was really weird, but it was a mistake in the `SecurityGuardTest` that I implemented, and it was even weird because it works in my PC xD. But basically there was a test that execute `clear()`, and it removed the `assert(boolean)` local method, and it cause some problems at some another tests 😅.

If there is some issue yet, I can see it in the next weekend probably, Good Night ✌️

src/main/java/bsh/security/SecurityError.java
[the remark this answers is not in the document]
nickl-: The thing is once it's public it must stay public and everything exposed needs to be maintained as if it is used publicly. Can investigate when `UtilEvalError` turned public, not sure it comes with a satisfying why though.

Prefer not to expose classes if we can help it, to save on maintenance.

src/main/java/bsh/security/SecurityGuard.java
nickl-: Makes sense.

nickl-:
> > Please also see Scrutinizer exceptions to rectify.
> 
> Abou this, I fixed the code style issues of Scrutinizer.
> 
> The tests results was really weird, but it was a mistake in the `SecurityGuardTest` that I implemented, and it was even weird because it works in my PC xD. But basically there was a test that execute `clear()`, and it removed the `assert(boolean)` local method, and it cause some problems at some another tests 😅.
> 
> If there is some issue yet, I can see it in the next weekend probably, Good Night ✌️

Yes these things can be a pain, thank you.

src/main/java/bsh/security/MainSecurityGuard.java
nickl-: Ok I see what you mean

nickl- (approved):

Net-0:
Hello @nickl- , I'm improving the code coverage, I really forget to do some tests 😅.

I also make some code improvements while doing the new tests for `SecurityGuardTest`:
- Unwrap the Primitive values in `MainSecurityGuard` before give them to the `SecurityGuard`, to the logic validate the "pure" value.
- Reflection to get the length of an Array will also trigger the `SecurityGuard#canGetField`

I'm going to see the other tests tomorrow, good night ✌️

Net-0:
I guess that I finished the tests, I made all missed tests for the implementation of SecurityGuard, but I ignored the files that have missed tests that the code wasn't implemented in this PR.

Can I run the codecov to see the code coverage or only you can ? I didn't find this option.... 🫠

After all, I guess that now everything looks good to me, but if it's not 😬, I probably can see any issues in the next weekend.

Good afternoon ✌️.

nickl-:
> Can I run the codecov to see the code coverage or only you can ? I didn't find this option.... 🫠

There is a code coverage in the maven build script too, although it is not the same as codecov, which is part of the automated build on github. It should run automatically when you push to this branch, that is the idea anyway, unless something went wrong with it.

nickl- (approved):
Waiting for code coverage

Net-0:
Hello,

I know that I'm going to say it again, but now I guess that everything is fine :D

About the changes, I basically make an adjust in the `bsh.BSHTypedVariableDeclaration` because when you declare a class field, it was getting the field value even though it being useless ( because the return value after evaluating a script when the last statement is a class declaration is the class itself ) and thus it was triggering the SecurityGuard to validate if can get that field. And I also used the jacoco that is in the maven scrips that you recommended, then I improved the tests and code coverage until I achieved the 100%.

But I guess that there is something weird, because I guess that codecov isn't updating:

codecov now:
image


jacoco results:
image
image

Good night and weekend ✌️, see you in the next weekend.

Net-0:
Everything seens nice, and I guess that there is nothing else that I can improve in this PR, so I'll just wait your review @nickl- 🙂 and approve or disapprove.

I also won't make any changes at this PR anymore, but I'm going to open a last PR that I guess that you'll like 😁

nickl-:
Not sure what is going on on codecov now but I am merging this, thank you for your patience and hard work.

Any additional problems can be handled with new PRs.

Issue #753, opened 2024-01-02
Problem handling multi-line comments


opeongo:
There seems to be a problem in the way that multi-line comments are handled in the grammar.  The regular expression token for the multi-line comment also catches formal comments, so in fact formal comments are never produced: they are always all matched and skipped by the MULTI_LINE_COMMENT token.

Here is the line in question from the .jjt file:
https://github.com/beanshell/beanshell/blame/b35aac408f9c86ec20f1201a5ae9582ed99e2ae2/src/main/jjtree/bsh.jjt#L312

This problem could be fixed with the following regular expression that would *not* gobble formal comments:
```java
    ("/***" (["*"])* | "/*" ~["*"]) (~["*"])* "*" ("*" | (~["*","/"] (~["*"])* "*"))* "/">
```

However, I don't think that this is the right fix.  Javadoc accepts multiline comments that start with two or more asterisks.  The existing regular expression will skip comments that start with three or more asterisks, which is different behaviour than Java.  So parsing Java and looking for formal comments is broken w.r.t. Java for syntactically correct Java code.  BeanShell claims that it can parse Java files, so this is a bit of a misconfiguration.

I think that the correct fix is to change the MULTI_LINE_COMMENT regular expression to only take ```/*``` comments as multi-line comments.  Any comment that begins with two or more asterisks should be passed on so that it can be parsed as a FORMAL_COMMENT.  In that case the FORMAL_COMMENT regular expression should become:
```java
 "/**" "*"* (~["*"])* "*" ("*" | (~["*","/"] (~["*"])* "*"))* "/"
```

opeongo:
When you pull on the string more issues crop up.  

Since FORMAL_COMMENT is defined as a regular token then it has to be included in to the grammar.  Currently a FORMAL_COMMENT may only appear where a block statement is expected.  This means that in BeanShell formal comments may not appear within statements.  So the following which is legal in Java is not allowed in BeanShell:
```java
if (1 == 2)
    assert(false);
/**
This comment is OK in java, not allowed in BeanShell
*/
else if (2 == 2)
    assert(true);
/**
Same with this one.
*/
else
    assert(false);
```

Maybe this seems like a corner case, but it will throw a parsing error.

Perhaps there needs to be some trick with lexical states to ignore formal comments within statements?

opeongo:
[...]

PR #754, opened 2024-01-03
Fix issue #753

opeongo:
- formal comments were always screened out, now are handled
- add test

### DIFF
--- a/src/main/jjtree/bsh.jjt
+++ b/src/main/jjtree/bsh.jjt
@@ -309,7 +309,7 @@ SPECIAL_TOKEN : /* COMMENTS */
  /* Moved FORMAL_COMMENT to a real token.  Modified MULTI_LINE_COMMENT to not
     catch formal comments (require no star after star) */
 | <MULTI_LINE_COMMENT:
-    ("/***" (["*"])* | "/*") (~["*"])* "*" ("*" | (~["*","/"] (~["*"])* "*"))* "/">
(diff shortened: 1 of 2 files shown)


nickl-:
Looks good

opeongo:
Need #732 merged before I can fix [bshdoc.bsh](https://github.com/beanshell/beanshell/blob/b35aac408f9c86ec20f1201a5ae9582ed99e2ae2/scripts/bshdoc.bsh).  It also has a problem with method lookups, where the interpreter has trouble distinguishing ```bshdoc(String)``` from ```bshdoc(String[])```.

nickl-:
> Need #732 merged before I can fix [bshdoc.bsh](https://github.com/beanshell/beanshell/blob/b35aac408f9c86ec20f1201a5ae9582ed99e2ae2/scripts/bshdoc.bsh). It also has a problem with method lookups, where the interpreter has trouble distinguishing `bshdoc(String)` from `bshdoc(String[])`.

Not sure how this relates to this issue, this PR is closed right?

opeongo:
The first version was a minimal fix that only worked with formal comments that had two **.   Other formal comment styles (three of more * or formal comments within statements) would be still be ignored or would cause parsing errors.

I have another version that fixes all the problems with parsing Java comments.  In this new patch BeanShell properly parse comments the same way as Java.  The consequence is dropping the BSHFormalComment node and adding a few accessor methods to obtain comments.

I updated bshdoc.bsh to use the new accessors, even though it is not being used in the build at the moment.  At least this should the new methodology for working with comments in the parser.

nickl-:
#756 has been reviewed

PR #756, opened 2024-01-08
Better fix for #753

opeongo:
- All comments are special tokens
- remove BSHFormalComment and add accessor methods to Parser
- add and modify tests
- adjust bshdoc.bsh to use accessor methods


scripts/bshdoc.bsh
```diff
@@ -147,8 +149,9 @@ bshdoc( filename )
 /**
     bshdoc list comment.
 */
-bshdoc( String [] filenames )
+bshdocl( String [] filenames )
 {
+System.err.println("filenames="+Arrays.toString(filenames));
```
nickl-, line 154: Is this just for debugging?

scripts/bshdoc.bsh
```diff
@@ -170,4 +173,5 @@ if ( list.length == 0 ) {
     print("usage: bshdoc file [ file ] [ ... ]");
     return;
 }
-bshdoc( list );
+System.err.println("list="+Arrays.toString(list));
```
nickl-, line 176: Same as previous instance of `System.err.println`

src/main/jjtree/bsh.jjt
```diff
@@ -1283,18 +1327,6 @@ void BlockStatement() :
 |
   // Allow BeanShell package declarations in any block
   PackageDeclaration()
-|
-  FormalComment()
-}
-
-void FormalComment() #FormalComment( retainComments ) :
```
nickl-, line 1290: Ok so we are turning FormalComment into a Node and not a Token anymore.

Not sure I follow the reasoning here, please explain.

nickl- (changes requested):
Looks like there is still debugging code in here

src/main/jjtree/bsh.jjt
nickl-: My bad

nickl-:
Looks good, just check on the debugging code.

nickl-:
Still waiting on review fixes

opeongo:
Updated now for review comments

jimjag:
Concept is OK but we've diverged enough that any implementation would require some rework. Closing this PR and recommending a new one if this is still something to be considered with a new patch.

PR #757, opened 2024-01-08
Add jdk 21 to the test matrix

opeongo:
As the title says, include jdk21 (a long term support release) to the test matrix


nickl-:
Awesome and it appears all is still working so RTBC

PR #758, opened 2024-01-22
EvalError vs dynamic language exception causes #556

opeongo:
This is a fix for issue #556

Catch evaluation errors that are caused by common scripting errors: syntax, name lookup failures, etc.  Allow the BeanShell script authors to catch these exceptions and attempt recovery or graceful shutdown.

The new class EvalException is a subclass of EvalError.  This will be used by benign exceptions where the interpreter can simply unwind from the point of error and carry on in the catch or finally block.


nickl-:
This was a huge help, thank you @opeongo

Issue #747, opened 2023-06-29
Arrays of boxed primites should not be initialized


opeongo:
The following code defies all expectations and breaks code that expects Java semantics:
https://github.com/beanshell/beanshell/blob/5e54a86566ff81e20af75cb221d534ba956a7709/src/main/java/bsh/BSHAllocationExpression.java#L305

Can you explain the rational for initializing arrays with default values?  BeanShell is a Java-type language, and this is so different and so unexpected.  Why do this?

opeongo:
Not trying to be confrontational, but this is so different from expectations that I don't see how this fits with BeanShell/Java model.  I really think it is worth revisiting this one.

opeongo:
I added a PR to fix this issue.  Discussion welcome.

nickl-:
If I understand you correctly you want arrays declared as box typed variables, to initialize with null as default value instead of their primitive default values?

Can't remember the reasons now, but it does seems intuitive since we auto box and unbox, and auto narrow and widen numbers.

Need more time to consider this.

nickl-:
It allows us to do things like:

```java
BeanShell 3.0.0-SNAPSHOT.5569
bsh % Integer[] bi = new int[5];
--> $1 = {0I, 0I, 0I, 0I, 0I} :Integer[]
bsh % int[] pi = new Integer[5];
--> $2 = {0I, 0I, 0I, 0I, 0I} :int[]
```

nickl-:
Ok still works with #747

```java
BeanShell 3.0.0-SNAPSHOT.5570
bsh % new Integer[5];
--> $0 = {null, null, null, null, null} :Integer[]
bsh % int[] pi = new Integer[5];
--> $1 = {0I, 0I, 0I, 0I, 0I} :int[]
```

There must've been a reason for doing this, it will come to me.

nickl-:
I agree, the BeanShell philosophy is "Works like JAVA but does not break like JAVA", null initialization is not considered a break.

opeongo:
> I agree, the BeanShell philosophy is "Works like JAVA but does not break like JAVA", null initialization is not considered a break.

So, you agree that null is not zero?  Or you agree something else?

This is broken, it should throw an exception rather than changing semantics:
```java
> BeanShell 3.0.0-SNAPSHOT.5570
> bsh % new Integer[5];
> --> $0 = {null, null, null, null, null} :Integer[]
> bsh % int[] pi = new Integer[5];
[...]

PR #760, opened 2024-01-23
Undo instantiation of boxed priimitie object arrays with default values

opeongo:
This is a fix for issue #747.


nickl-:
Merged as per #747

Issue #767, opened 2024-07-29
Float data types are unnecessarily widened, causing problems


opeongo:
There seems to be a problem with float data types always being widened as a result of an expression, regardless of the need to do so.  For example:
```java
float a = 1;
float b = 2;
c = a * b;
print(c);
2.0d :double
```

This shows up as a problem when the result of the expression is used as an argument to a method that is expecting the narrower data type.  The outcome is an exception that the method matching the signature of the wider variable cannot be found.  For example:
```
Float.isInfinite(1f*2f);
Sourced file: inline evaluation of: ``Float.isInfinite(1f*2f); '' : Error in method invocation: Static method isInfinite(double) not found in class'java.lang.Float' : at Line: 1 : in file: inline evaluation of: ``Float.isInfinite(1f*2f); '' : Float .isInfinite ( 1f * 2f ) 

Caused by
    bsh.ReflectError: Static method isInfinite(double) not found in class'java.lang.Float'
```

The [internal documentation for the Operators.promotePrimatives](https://github.com/beanshell/beanshell/blob/3c576079f9395bd9c53af75c0224b6831c84aa70/src/main/java/bsh/Operators.java#L509) method says
```
        Promote the pair of primitives to the maximum type of the two.
        e.g. [int,long]->[long,long]
```
but it appears that floats are unconditionally promoted to doubles by this method.  Even if both operands are float, they will both be promoted to double.

I would like to change this behaviour to be more like what the above comment says:  only widen as much as required.  This would allow the expression 
```
Float.isIninifte(1f*2f);
```
to work as expected.

I notice that the [Primitive.shrinkWrap method](https://github.com/beanshell/beanshell/blob/3c576079f9395bd9c53af75c0224b6831c84aa70/src/main/java/bsh/Primitive.java#L182) also does an unconditional promotion of float to double.  This means that there are a few places in the code base that follow a policy of widening floats to doubles even when not required.  Will there be an impact on other parts of the code it I put together a PR to make this change?

nickl-:
I believe we will have similar issues with byte, short, and char (since we can also treat it as a number and perform math operators).

Under Operators it seems ok to use the larger types to perform the operations, what we need to do is ensure that we can shrink to lower types as needed.

Interesting problem, well spotted.

opeongo:
Java always promotes byte, short and char arithmetic to int, so these are already being handled correctly.

Issue #71, opened 2018-06-18
Numeric Types


nickl-:
Fixed numeric types - see beanshell/beanshell@4f2e0e2f

Attempted a new approach, the idea was to start at the smallest data type, `byte` and then seamlessly widen the range as required. `byte`, `short`, `int`, `long`, `float`, `double` It worked great from within beanshell but things go pear shape when interacting from java who believes numbers are `int`. If we only had access to primitives it would be fine because `int == byte` is happiness but `java.lang.Integer != java.lang.Byte` not so much joy. Still the premise works so instead of fighting against the stream lets stick with starting at `int`.

We needed to include `BigInteger` and `BigDecimal` I did several bench marks and things have improved a lot but the primitives are still easily 20 times faster in every aspect, this is rather significant. The roll out thus went as follows.

Not floating points

 * Integer
 * Long
 * BigInteger 

Floating points

 * Float
 * Double
 * BigDecimal

When we are crunching we use the biggest primitives first `long` and `double` and overflowing into `BigInteger' and `BigDecimal` respectively. When computation is done we shrink down to the lowest capable denominator for the result. This also allows us to cast down (narrowing) allowing things like:

```java
int i = 2L; // result 2 : int
float f = 2.0; // result 2.0 : float
double d = 2.0; // no magic
f *= d; // result 4.0 : float
``` 

Loos typed local variables will automatically widen but we maintain type safety for type declared variables and should you go out of range, the usual error messages will prompt you that you cannot assign `long` to `int` etc.

We still only have the default java type suffixes implemented, which are respected, but it may be handy to specify others so that you can express what you want.

The java type suffixes:

type|suffix
-- | --
long | L / l
float | f / F
double | d / D

Groovy adds the following:

type|suffix
-- | --
BigInteger | G / g
Integer | I / i
BigDecimal | G / g

I would like to go even further and include `byte` and `short` as well.

Type | Suffix
-- | --
BigInteger | ? / ?
Long | L / l
Integer | I / i
Short | S / s
Byte | B / b
BigDecimal | ? / ?
Double | D / d
Float | F / f

Not sure what to call the `Big*` types the `G` obviously has domain connotations. Ideas? 

Implemented all the operators and supplied unit tests, also added  the power operator `**` and power assign `**=`, available in groovy/python/etc while we were at it. Are there any others we should consider? What about something for regex?

Refactored Primitives and operators code, there is still a whole lot of duplication so we could probably take this through a few more iterations. 

[...]

PR #768, opened 2024-07-29
Fix problem of unnecessarily widening float values

opeongo:
This is a fix for issue #767 where float expressions are widened to doubles.

### DIFF
--- a/src/main/java/bsh/Operators.java
+++ b/src/main/java/bsh/Operators.java
@@ -570,28 +570,28 @@ class Operators implements ParserConstants {
         Number lnum = promoteToInteger(lhs);
         Number rnum = promoteToInteger(rhs);
 
-        if (lhs instanceof BigDecimal) {
-            if (!(rhs instanceof BigDecimal))
-                rhs = Primitive.castNumber(BigDecimal.class, rnum);
-        } else if (rhs instanceof BigDecimal) {
-            lhs = Primitive.castNumber(BigDecimal.class, lnum);
-        } else if (lhs instanceof Float && rhs instanceof Float) {
-            ; // do nothing
-        } else if (Types.isFloatingpoint(lhs) || Types.isFloatingpoint(rhs)) {
-            if (!(lhs instanceof Double))
-                lhs = Double.valueOf(lnum.doubleValue());
-            if (!(rhs instanceof Double))
-                rhs = Double.valueOf(rnum.doubleValue());
-        } else if (lhs instanceof BigInteger) {
-            if (!(rhs instanceof BigInteger))
-                rhs = Primitive.castNumber(BigInteger.class, rnum);
-        } else if (rhs instanceof BigInteger) {
-            lhs = Primitive.castNumber(BigInteger.class, lnum);
-        } else {
-            if (!(lhs instanceof Long))
-                lhs = Long.valueOf(lnum.longValue());
-            if (!(rhs instanceof Long))
-                rhs = Long.valueOf(rnum.longValue());
+        if (!(lhs instanceof Float && rhs instanceof Float)) {
+            if (lhs instanceof BigDecimal) {
+                if (!(rhs instanceof BigDecimal))
+                    rhs = Primitive.castNumber(BigDecimal.class, rnum);
+            } else if (rhs instanceof BigDecimal) {
+                lhs = Primitive.castNumber(BigDecimal.class, lnum);
+            } else if (Types.isFloatingpoint(lhs) || Types.isFloatingpoint(rhs)) {
+                if (!(lhs instanceof Double))
+                    lhs = Double.valueOf(lnum.doubleValue());
+                if (!(rhs instanceof Double))
+                    rhs = Double.valueOf(rnum.doubleValue());
+            } else if (lhs instanceof BigInteger) {
+                if (!(rhs instanceof BigInteger))
+                    rhs = Primitive.castNumber(BigInteger.class, rnum);
+            } else if (rhs instanceof BigInteger) {
+                lhs = Primitive.castNumber(BigInteger.class, lnum);
+            } else {
+                if (!(lhs instanceof Long))
+                    lhs = Long.valueOf(lnum.longValue());
+                if (!(rhs instanceof Long))
+                    rhs = Long.valueOf(rnum.longValue());
+            }
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/bsh/Operators.java
```diff
@@ -399,9 +401,62 @@ static Object bigIntegerBinaryOperation(BigInteger lhs, BigInteger rhs, int kind
...
+            case LSHIFTX:
+            case RSIGNEDSHIFT:
+            case RSIGNEDSHIFTX:
+            case RUNSIGNEDSHIFT:
+            case RUNSIGNEDSHIFTX:
+                throw new UtilEvalError("Can't shift float values");
+
+        }
+        if ( OVERFLOW_OPS.contains(kind) )
+            return bigDecimalBinaryOperation(BigDecimal.valueOf(lhs), BigDecimal.valueOf(rhs), kind);
```
nickl-, line 451: I wonder if we shouldn't first overflow to double here

src/main/java/bsh/Operators.java
```diff
@@ -399,9 +401,62 @@ static Object bigIntegerBinaryOperation(BigInteger lhs, BigInteger rhs, int kind
...
+                return lhs / rhs;
+
+            case MOD:
+            case MODX:
+                return lhs % rhs;
+
+            case POWER:
+            case POWERX:
+                double check = Math.pow(lhs, rhs);
+                if ( Double.isInfinite(check) )
```
nickl-, line 436: Should we not try and return a float here?

src/main/java/bsh/Operators.java
```diff
@@ -515,26 +570,28 @@ static Object[] promotePrimitives(Object lhs, Object rhs)
...
-            if ( !(rhs instanceof BigInteger) )
-                rhs = Primitive.castNumber(BigInteger.class, rnum);
-        } else if ( rhs instanceof BigInteger ) {
-            lhs = Primitive.castNumber(BigInteger.class, lnum);
-        } else {
-            if ( !(lhs instanceof Long) )
-                lhs = Long.valueOf(lnum.longValue());
-            if ( !(rhs instanceof Long) )
-                rhs = Long.valueOf(rnum.longValue());
+        if (!(lhs instanceof Float && rhs instanceof Float)) {
```
nickl-, line 573: It is hard to see what happens here but this seems wrong.

From the existing function:
https://github.com/beanshell/beanshell/blob/53b0436f4e2777e5ec8c8f60e565c7ef48f3b18b/src/main/java/bsh/Operators.java#L513-L541

There is a precedence BigDecimal, Double, BigInteger, Long and it seems this check needs to be  around line 524 when we are checking for Doubles.

src/main/java/bsh/Primitive.java
```diff
@@ -183,6 +183,8 @@ public static Primitive shrinkWrap(Object number) {
         if (!(number instanceof Number))
             throw new InterpreterError("Can only shrink wrap Number types");
         Number value = (Number) number;
+        if (value instanceof Float)
+            return new Primitive(value.floatValue());
```
nickl-, line 187: I think this needs to do something similar to int and long, comparing the values instead of the type.

See full function 193 and 195:
https://github.com/beanshell/beanshell/blob/53b0436f4e2777e5ec8c8f60e565c7ef48f3b18b/src/main/java/bsh/Primitive.java#L178-L198

I wonder if this would not be sufficient, to just shrink the double result to float if it fits?

nickl- (changes requested):
Needs work.

I'm wondering if there isn't a simpler solution, to use the double functionality as is, but if it is floats we do a shrink (if they fit) else expand to Double/BigDecimal as needed. This will be a much smaller change and less things that can cause additional headaches.

src/main/java/bsh/Operators.java
```diff
@@ -190,6 +190,8 @@ static <T> Object binaryOperationImpl( T lhs, T rhs, int kind )
             return bigIntegerBinaryOperation( (BigInteger) lhs, (BigInteger) rhs, kind );
         if (lhs instanceof BigDecimal)
             return bigDecimalBinaryOperation( (BigDecimal) lhs, (BigDecimal) rhs, kind );
+        if (lhs instanceof Float)
+            return floatBinaryOperation( (Float) lhs, (Float) rhs, kind );
```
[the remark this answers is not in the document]
opeongo: binaryOperationImpl only has one usage (in Operators.binaryOperation()), and that is after lhs and rhs have been passed through Operators.promotePrimitves().   At this point if lhs is Float it is not possible for rhs to not be Float.

src/main/java/bsh/Operators.java
opeongo: Can we just stick with Java math please?  Some of us depend on it.

src/main/java/bsh/Operators.java
opeongo: Math.pow returns a double, so no one is expecting to get a float here.

src/main/java/bsh/Operators.java
opeongo: I can refactor this but the outcome will be the same.

src/main/java/bsh/Primitive.java
opeongo: If you do the widen float to double, do the operation, and then narrow you will get a different result than doing the operation in float.  Best to keep this the same as Java.

In Java byte and short use int as a carrier type.  Arithmetic on any combination of these types will yield an int.  Best to keep this the same as Java.

wilx:
Compatibility note for the follow-up prepared on top of this PR: the changes intentionally align float arithmetic with Java, including mixed float/integral promotion, rounding before assignment, signed zero, NaN, and infinity on overflow. Results are determined by operand types and Java arithmetic, rather than widened to preserve extra precision.

This deliberately revises the double-first arithmetic and float-overflow widening documented in #71. The `long += float` example in https://github.com/beanshell/beanshell/issues/71#issuecomment-404611375 also produces unchanged values in compiled Java: Java promotes the long operand to float before addition, then converts the result back to long.

The follow-up preserves the existing explicit BigInteger/BigDecimal and power extensions; broader double arithmetic remains outside this change. It is currently local and passes JDK 8 clean verification: 631 tests, 6 skipped, zero failures/errors and Checkstyle violations.

PR #772, opened 2024-08-31
Visibility issue of MainSecurityGuard

Net-0:
I just forgot to add a `public` in the class :|, sorry for that

### DIFF
--- a/src/main/java/bsh/BSHAllocationExpression.java
+++ b/src/main/java/bsh/BSHAllocationExpression.java
@@ -32,6 +32,8 @@ import java.lang.reflect.Array;
 import java.lang.reflect.InvocationTargetException;
 import java.util.concurrent.CompletionException;
 
+import bsh.security.SecurityError;
+
 /**
     New object, new array, or inner class style allocation with body.
 */
--- a/src/main/java/bsh/Interpreter.java
+++ b/src/main/java/bsh/Interpreter.java
@@ -41,6 +41,8 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.util.ResourceBundle;
 
+import bsh.security.MainSecurityGuard;
+
 /**
     The BeanShell script interpreter.
 
--- a/src/main/java/bsh/Name.java
+++ b/src/main/java/bsh/Name.java
@@ -812,8 +812,8 @@ class Name implements java.io.Serializable
                 Object instance = classNameSpace.getClassInstance();
                 Class<?> classStatic = classNameSpace.classStatic;
 
-                // Validate if can invoke this super method
-                Interpreter.mainSecurityGuard.canInvokeSuperMethod(instance.getClass().getSuperclass(), instance, methodName, args);
+                // Validate if can invoke this method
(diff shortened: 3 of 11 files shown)


Net-0:
doc: https://github.com/beanshell/beanshell/wiki/SecurityGuard

Net-0:
Hello @nickl- , I created this PR to fix the visibility of MainSecurityGuard class, but reading the code, I guess that makes sense that the SecurityGuard classes be in a package called `security` at all, it makes the code less messy.

About the `bsh.Types` visibility, I added a public method in `bsh.Reflect` of `getType` to solve that.

I also make a doc and already put the link above, it seems good. When you aprove the PR, I could put the link in the main wiki page for the SecurityGuard wiki page.

It seems ok, I'm going to finish my PR with generated class improvements to I be finally able to end the lambda support, see ya :D

wilx:
LGTM. I verified external-package guard registration and ran JDK 8 clean verify on head 68d0e726: 759 tests, 6 skipped, zero failures/errors, and zero Checkstyle violations.

jimjag:
Thanks for the PR, but this is more than a visibility fix — worth flagging before merge.

**Scope**: this moves `MainSecurityGuard`, `SecurityGuard`, and `SecurityError` from `bsh` to a new `bsh.security` package and widens most of their methods to `public`. That's a real API change for anyone referencing the old `bsh.*` FQNs, not just an access-modifier tweak.

**Behavior change**: `MainSecurityGuard.canInvokeSuperMethod(superClass, instance, methodName, args)` — the dedicated hook a custom `SecurityGuard` could use to restrict `super.foo()` calls separately from ordinary method calls — has been removed from the interface and implementation. The `super.foo()` call site in `Name.java` now goes through the generic `canInvokeMethod(instance, methodName, args)` instead, using the subclass instance rather than the actual superclass being invoked.

Any existing `SecurityGuard` implementation that overrides `canInvokeSuperMethod()` to enforce super-call-specific restrictions will silently stop having that check enforced. This shows up in the PR's own test changes: `cant_invoke_super_method` used to assert `"Can't invoke this super method: Cls1.method()"` and now asserts `"Can't invoke this method: Cls2.someMethod()"` — the test was adjusted to match the new (different) behavior rather than preserving equivalent coverage.

Could you either restore `canInvokeSuperMethod` (or an equivalent super-call-specific hook), or, if consolidating super-calls into the generic `canInvokeMethod` check is intentional, call that out explicitly in the PR description so it can be evaluated as a security-model change rather than a "forgot a public" fix?

(`Types.prettyName()` removal looks fine — its only caller was updated inline to `Class#getTypeName()` with equivalent output.)

jimjag:
Correction to my earlier comment: I overstated the impact. SecurityGuard hasn't been in a release, and on current master `MainSecurityGuard` and `SecurityError` are package-private, so code outside the `bsh` package can't call `Interpreter.mainSecurityGuard.add(...)` — which is exactly what this PR fixes. There are no existing external guards that would lose a `canInvokeSuperMethod` check, and the package move doesn't break any released API.

The only open question is a design one: is replacing the super-call-specific hook with `canInvokeMethod` (checked against the subclass instance) intentional? If so, a line in the PR description saying so is all that's needed.

PR #788, opened 2026-09-07
Update bundled ASM to 9.10.1

wilx:
Update bundled ASM from 6.2 to 9.10.1 using the complete upstream core sources, including the class reader needed for the writer's long-jump rewriting. A scripted BeanShell class with 96 overloaded constructors exercises this path: the old bundled copy drops the generated constructors, causing instantiation to fail with `Can't find default constructor`. The regression in `ClassGeneratorTest` evaluates that script and checks both constructor chaining and direct instantiation.

Keep the `bsh.org.objectweb.asm` relocation and Java 8 bytecode target. Six `@Deprecated(forRemoval = false)` annotations are reduced to `@Deprecated` so the sources also compile on JDK 8. The bundled-source README records the upstream archive, SHA-256, and exact source transformations.

Validation:

- `mvn -B -ntp clean verify` on Java 8: 695 tests, 6 skipped, no failures or errors; no Checkstyle violations.
- Confirmed the BeanShell script fails against unmodified upstream master and succeeds with the update.
- `git diff --check` passes.

### DIFF
--- a/src/main/java/bsh/org/objectweb/asm/AnnotationVisitor.java
+++ b/src/main/java/bsh/org/objectweb/asm/AnnotationVisitor.java
@@ -1,0 +1,168 @@
+// ASM: a very small and fast Java bytecode manipulation framework
+// Copyright (c) 2000-2011 INRIA, France Telecom
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+// 1. Redistributions of source code must retain the above copyright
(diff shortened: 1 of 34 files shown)


jimjag:
Merged the ASM 9.10.1 update (3f3a26e4) to master, thanks. I checked the bundled sources against the upstream `asm-9.10.1-sources.jar`: the checksum and the three documented replacements match exactly.

I left out the second commit: its CI and README changes conflict with the build restructuring from #791 (JDK 17 build, tests run on each matrix JDK). Java 25 has been added to the new CI matrix in a separate commit.

PR #789, opened 2026-09-07
Fix namespace caching and defer member-cache initialization

wilx:
Builds on #741 to address the namespace caching failures discussed in #659. This preserves the original PR's commits, including proper namespace-key equality, synchronous weak/soft value caching, and enhanced-for-loop namespace reuse.

Changes on top of #741:

- Preserve a live replacement when processing an older queued reference. Previously, `ValueReferenceMap.clean()` removed entries by key alone, so cleanup could evict a replacement created before the old reference was processed. Removal now checks both the key and the queued reference.
- Defer member-cache construction until reflection metadata is requested. Previously, a class-literal lookup could throw `NoClassDefFoundError` when an unused method referenced an unavailable dependency, because `cacheClassInfo()` synchronously enumerated the class's members. Class lookup now records the class without eagerly building its member cache.

Add deterministic regression tests for weak and soft reference replacement, plus a class-loading fixture with an unavailable method dependency. The cleanup tests explicitly clear and enqueue references rather than relying on GC timing. All three new regression tests were confirmed to fail against #741's original implementation and pass with these fixes.

Validation:

- `mvn -B -ntp clean verify` on JDK 8: 615 tests, 6 skipped, no failures or errors; no Checkstyle violations.
- Focused cache and class-manager tests: 12 tests, all passed.
- `git diff --check` passes.


opeongo:
Where is the PR for cd0184ebb31175a993ed155148fec87eb03c9fb9 (Restore weak/soft key semantics in ValueReferenceMap).  I would like to take a closer look at this one but I don't think I have the right context for the changes.

PR #794, opened 2026-09-08
Allow formal comments within statements and preserve bshdoc output

### LINKED ISSUE #753: Problem handling multi-line comments

wilx:
Formal comments can currently break valid code when placed inside a statement, for example between an `if` branch and its `else`. Incorporate #756's change to treat comments as special tokens, while preserving documentation produced by `bshdoc`.

Based directly on #756's head `cdddd915ca0a1b93213472a5a119ef49ad3f6723`, retaining both original commits and their authorship. Three separate follow-up commits address the review findings:

- `7a8a1ce0`: Preserve documentation in comment-only files and trailing comments attached to EOF. Add `Parser.getFormalCommentsBeforeToken(Token)` and use it after parsing; retain existing file headers.
- `3f8f981b`: Preserve the first file comment separately from the last method comment, including adjacent comments with identical text. Avoid duplicating a single method comment as file documentation.
- `5c8fecf3`: Remove the two outstanding stderr debug prints and their unused import.

Add ten tests: nine run the actual `bshdoc` CLI in a separate JVM and inspect its XML output, including multi-file output and empty stderr; one checks EOF comment ordering and filtering. The EOF, adjacent-comment, and debug-output regressions were each confirmed to fail before their respective fixes.

Validation on JDK 8:

- `mvn -B -ntp clean verify`: 627 tests, 6 skipped, zero failures/errors, zero Checkstyle violations.
- All 96 comment-placement cases pass; #756's base and current master pass 65 of them. The completed branch also passes 20 direct comment-accessor checks.
- `git diff --check` and ancestry checks pass.

These results cover #756's tip plus the three follow-up commits; master has not been merged into this branch, and hosted CI completion is not yet verified.

Builds on #756.
Fixes #753.


jimjag:
I merged current master into this branch locally (no conflicts). On JDK 21, everything passes except the nine CLI tests in `BshDocTest`, and the cause is a JDK 9+ issue rather than the comment handling.

`BshDocTest` runs `bshdoc` in a child JVM using the test JVM's `java.home`. `scripts/bshdoc.bsh` line 43 calls `setAccessibility(true)`, which on newer JDKs throws:

```
java.lang.reflect.InaccessibleObjectException: Unable to make field private final byte[] java.lang.String.value accessible: module java.base does not "opens java.lang" to unnamed module
```

`bsh.Interpreter` reports "Script threw exception" but still exits 0, so `render()` gets empty output and each test fails with `SAXParseException: Premature end of file`. Running the same compiled classes by hand gives correct XML on JDK 8 and this exception on JDK 23.

Master's CI now builds with JDK 17 and runs tests on 8, 11, 17 and 21, and JDK 17+ enforces this module restriction by default, so the 17 and 21 jobs should fail once this is rebased. I haven't checked JDK 11. That `setAccessibility(true)` line dates from 2012, so `bshdoc` itself probably hasn't worked on JDK 9+ for a long time; these tests are just the first to exercise it. Options: pass the needed `--add-opens` flags to the child JVM, or change `bshdoc.bsh` so it doesn't need accessibility.

Minor: renaming `bshdoc(String[])` to `bshdocl` is outside the comment fix. Fine if intended, but worth a line in the description.

wilx:
I have rebased this branch. I have removed the unnecessary privileged access from bshdoc. The rename to bshdocl is not mine. It was in the original PR. I have updated JaCoCo for Java 25 support. Everything seems to build fine now.

PR #852, opened 2026-09-18
Fix message-only ParseException losing its message

### LINKED ISSUE #839: A message-only ParseException loses its message and throws from getErrorLineNumber()
```
private public int x = 1;
```
is correctly rejected, but through `ParseException(String)`, whose `currentToken` is null. `getMessage()` ignores the actual stored message in that case and prints only "Unable to parse code syntax. Encountered:" with nothing useful after it, and `getErrorLineNumber()` throws a `NullPointerException` (around `ParseException.java:217`) instead of returning something sensible.

jimjag:
A `ParseException` built from a plain message — e.g. a modifier-conflict error like
`private public int x = 1;` — lost that message entirely. `ParseException` overrides
`getMessage()` to build a token-derived string, but when the exception comes from the message-only
constructor there's no token, so the override fell through to a bare
"Unable to parse code syntax. Encountered:" with nothing after it, and `getErrorLineNumber()`
threw an NPE dereferencing the same null token instead of returning something sensible.

Fix: `getMessage()` returns the exception's own stored message when there's no token to derive one
from, `getErrorLineNumber()` returns -1 in that case (matching `EvalError`'s contract), and a
null-guard closes a second latent NPE in the same message-building code that only shows up with
`Interpreter.DEBUG` on. Ordinary parse errors raised by the generated parser, which do carry a
token, are unchanged — pinned by a new regression test alongside the existing one.

Fixes #839.


jimjag:
@copilot resolve the merge conflicts in this pull request

PR #855, opened 2026-09-18
Fix Java-inapplicable fixed overload preempting varargs

### LINKED ISSUE #849: A Java-applicable varargs overload can be preempted by a Java-inapplicable fixed-arity one
```java
static String pick(boolean value, Runnable action) { return "fixed"; }
static String pick(String value, Runnable... actions) { return "varargs"; }
pick("x", () -> {});
```
Real javac resolves this to `pick(String, Runnable...)` — `"x"` has no legal conversion to `boolean` in real Java at any applicability phase, so the fixed-arity overload is never a candidate and phase 3 (varargs) is reached. BeanShell instead resolves to `pick(boolean, Runnable)`, because its overload-resolution machinery (both the lambda-aware path and the ordinary non-lambda path share this structure) tries every fixed-arity candidate across all four of its own assignability rounds to exhaustion before considering any varargs candidate at all — and bsh's own loosest round accepts conversions, like `String` to `boolean`, that have no equivalent in real Java.

This is trickier than just interleaving the two candidate lists by round: bsh's rounds don't cleanly separate "real Java conversions" from "bsh-only loose conversions" — a real, legal Java conversion (unboxing then widening, e.g. `Integer` to `long`, JLS 5.3) currently only matches in bsh's *last* round too, alongside conversions Java doesn't allow at all, so round number alone can't distinguish them. A correct fix likely needs bsh's earlier assignability rounds to recognize unbox-then-widen directly, ahead of any fix to how fixed-arity and varargs candidates compete.

Audgui-Byte:
Fixes #849.

A `boolean, Runnable` fixed-arity overload currently beats a Java-applicable `String, Runnable...` overload for `pick("x", () -> {})`, because BeanShell exhausts its permissive fixed-arity conversion round first. Resolve Java fixed and expanded-varargs candidates before trying BeanShell-only conversions, for both scripted and compiled methods and for ordinary and lambda arguments. The raw-lambda fallback remains last.

Java's unboxing followed by primitive widening is recognized in the earlier boxing round, so a legal fixed-arity `long` overload still beats varargs for an `Integer`. Widening uses the explicit Java conversion relationships; `Byte`/`Short` to `char` must not become Java-applicable.

Regression tests cover lambda and ordinary calls and legal/illegal wrapper conversions, including compiled and scripted overloads.

Validation on Windows / JDK 21:
- Before the fix, both issue-specific regression tests fail; the existing 68 tests in that class pass.
- After: the full suite completes with 1,254 tests, 13 skipped, no failures or errors, plus 4 passing shared-cache tests. The focused resolution class has 72 passing tests.
- `mvn -DskipTests verify` passes, with zero Checkstyle violations.
- Separate `javac` examples confirm the issue and wrapper-conversion expectations. An independent check of 64 wrapper-to-primitive combinations matches `javac` on both compiled and scripted calls.

The full suite was run from an ASCII-only worktree because the forked JaCoCo JVM fails to resolve the local non-ASCII workspace path.

Implementation and review used AI coding agents, with tests executed locally. Older JDKs and performance have not been retested.


jimjag:
Closing in favor of #859, which fixes this same issue (#849) and has been merged.

PR #865, opened 2026-09-21
Run scripted interface default methods against the implementing instance

### LINKED ISSUE #832: Scripted interface default methods fail when invoked through a plain implementing class
A scripted interface's default method that calls the interface's own abstract method fails when invoked on a plain scripted class implementing that interface — e.g.
```
interface Foo { int bar(); default int baz() { return bar() + 1; } }
class Impl implements Foo { int bar() { return 1; } }
Foo f = new Impl();
f.baz();
```
Assigning the implementation to the interface type (`Foo f = new Impl();`) also fails on its own. Exact error text wasn't captured when this was found — worth re-running to pin down the failure mode before fixing.

SulimanAbdulrazzaq:
Fixes #832.

A scripted interface's default method runs its body on the interface's static `This`, so a default method that uses the instance fails. The issue's example

```java
interface Foo { int bar(); default int baz() { return bar() + 1; } }
class Impl implements Foo { int bar() { return 1; } }
Foo f = new Impl();
f.baz();
```

fails twice on `main`:

- `Foo f = new Impl();` aborts with `Generated class: Error getting This bsh.ReflectError: No such field: _bshThisFoo for class: Foo`. `Types.castObject` asks every generated target type for its instance `This`, and an interface has none, so this breaks any assignment, cast, parameter or field typed as a scripted interface, with or without default methods.
- `baz()` throws `Cannot reach instance method: bar() from static context: Foo`. `ClassGeneratorUtil.generateMethod` compiles every non-static interface method to call `invokeMethod` on the interface's static `This`, so the body never sees the instance. A default method calling another default method fails the same way.

Changes:

- `ClassGeneratorUtil.generateMethod`: a default method still loads the interface's static `This`, but now also passes `this` and calls the new `This.invokeDefaultMethod`.
- `This.invokeDefaultMethod`: finds the method in the interface's static namespace as before, then runs it with its local namespace parented on a new namespace that imports the instance (`isClass`, `setClassInstance`), the way a class instance method runs. It uses the existing `parentNameSpace` argument of `BshMethod.invoke`. `bar()`, `this.bar()` and other default methods resolve on the instance with normal virtual dispatch, and interface constants and static methods still resolve through the interface.
- `Types.castObject`: a generated interface target keeps the value itself instead of looking up an instance `This`.

Left as is: calling a default method statically (`AAC.def(1)` in `InterfaceMethodsTest`) still fails; the check in `BshLambda` that rejects lambdas for scripted interfaces with default methods is unchanged; and an anonymous implementation (`new Foo() { ... }`, a proxy) still can't call a default method it doesn't override, which is a separate path.

Tests: four new cases in `InterfaceMethodsTest`: the issue's example, dispatch to an overriding subclass, `this.bar(x)` plus a default method calling another default method, and a class instance assigned, cast and passed as a scripted interface type. They error on `main` and pass with this change. `mvn -B clean verify` passes with a JDK 17 build and test JVMs 8, 11, 17, 21 and 25.


jimjag:
Thanks for this, the diagnosis of the two failures is right and reusing the `parentNameSpace` argument of `BshMethod.invoke` is the right shape. I built the branch and ran it: `mvn -o verify` passes, and the `abstract-method`, `this.bar()`, default-calls-default, inherited-interface, override-in-subclass and `Foo f = new Impl()` cases all work.

One regression to fix before merge. The wrapper namespace is `isClass` with the instance imported, so the implementing class's members are looked up before the interface's own constants and static methods. Java resolves names in a default method against the interface. Both of these worked on main (the static `This` was used) and now return the wrong value:

```java
interface A { int X = 10; default int getX() { return X; } }
class B implements A { int X = 99; }
new B().getX();   // main: 10, this PR: 99

interface A { static int h() { return 1; } default int callH() { return h(); } }
class B implements A { int h() { return 2; } }
new B().callH();  // main: 1, this PR: 2
```

The interface's static namespace needs to take precedence over the instance's imports for those two cases, while `bar()` and `this.bar()` still dispatch on the instance.

Could you also add tests for:

- the two cases above
- an implementing class that overrides a default method
- a default method on an enum implementing the interface, and one inherited through an interface that extends another scripted interface
- reading and writing an instance field via `this` in a default method (this works today, it just isn't pinned)

The branch is 14 commits behind `main` and `ClassGeneratorUtil.java` changed in that range. It merges cleanly, but please rebase and re-run the tests. A short entry in `CHANGES.md` for the user-visible fix would be good too.

Not for this PR: `new Foo() { ... }` still can't call an un-overridden default, as you noted.

SulimanAbdulrazzaq:
Thanks for the review. The branch is now rebased onto current main (49e4ca3):

- `3495cd5` Run scripted interface default methods against the implementing instance: the original commit, rebased. The one conflict was in `Types.castObject`, where main now uses `BshLambda.isWrapperClass`. The condition keeps that and adds `!toType.isInterface()`. `ClassGeneratorUtil.java` merged without changes.
- `4046e28` Resolve a default method's names against its interface first: fixes the regression.
  - A name in a default method now resolves as on main: through the interface's static namespace, then the scopes around it. It reaches the instance only when nothing there matches.
  - The exception is the interface's abstract and default methods, declared or inherited. Those dispatch on the instance, so `bar()`, `this.bar()`, default-calls-default and overrides behave as before.
  - Your two examples return 10 and 1 again.
  - The same shadowing also affected script-level names. With `int limit = 7;` and a script method `twice(int)`, a default method returning `twice(limit)` gave 0 when the implementing class had its own `limit` and `twice`. It gives 14 again, as on main.
  - Each of these three cases has a test.
- `363e49a` Pin more scripted interface default method cases:
  - an implementing class overriding a default method;
  - a default method on an enum constant;
  - a default inherited through `interface Sub extends Base`;
  - reading and writing `this.count` from a default method.

  The enum and `Sub extends Base` tests also check that the interface constant wins over an instance field with the same name.
- `723cbd3` Add a CHANGES entry for scripted interface default methods.

Test results (JDK 11):
- `mvn -B clean verify`: BUILD SUCCESS, with no failures or errors and 0 Checkstyle violations.
  - `default-test`: 1375 tests, 16 skipped.
  - `shared-cache-concurrency`: 4 tests.
  - `class-lifecycle-concurrency`: 9 tests, 1 skipped.
- `InterfaceMethodsTest`: 23 tests, all pass.
- On the rebased first commit alone (without `4046e28`), the same test class has 5 failures:
  - your two examples (2 instead of 1, and 99 instead of 10);
  - the script-level names test (0 instead of 14);
  - the enum test (105 instead of 16);
  - the extended-interface test (10200 instead of 1122).
- On main, the three name-resolution tests pass. The 8 tests that need the instance error with one of:
  - `Cannot reach instance method: ... from static context`
  - `No such field: _bshThisFoo for class: Foo`
  - `illegal use of undefined variable` for `this.count`

Left as is: a loosely typed field in the implementing class (`count = 5;` rather than `int count = 5;`) still can't be reached through `this` from a default method. It fails the same way on main.

Ready for another look when you have time.

